PR #48, opened 2016-02-28
Adds randomization in testint order

### LINKED ISSUE #9: Randomize unittest order

aopincar:
This patch adds an installation of 'pytest-random' PyPi package which
adds the ability to run the unittests in random order.
when running with '-v/--verbose' flag, the seed will be printed to
stdout to allow running in same order if want (--random-seed=xxxxx must
come with --random)

resolves #9


yfried-redhat:
You are trying to fix #52 in the same commit.
See #53 for an alternative

PR #56, opened 2016-02-28
Adds ir-installer entry point with ospd subparser

yfried-redhat:


cli/install.py
```diff
@@ -0,0 +1,209 @@
...
+    :param args: Namespace of argument parser
+    :returns: path to settings dir. if args, return path to parser settings
+        dir. If args.command0 exists, return path to subparser settings dir.
+    """
+    settings_dir = utils.validate_settings_dir(
+        CONF.get('defaults', 'settings'))
+    if args:
+        settings_dir = os.path.join(settings_dir,
+                                    ENTRY_POINT)
+    if hasattr(args, "command0"):
```
obaranov: Is the command0 component is required in the it the settings path? I'd suggest to handle situation when command0 is not defined (e.g. raise an exception).

yfried-redhat: It's a hack that should be handled by #41 
A proper Spec manager should provide an abstract spec class to work with

PR #68, opened 2016-02-29
Removes configure, Improves dict_merge, Adds merging unittest, Fixes Lookup unittests

### LINKED ISSUE #45: Remove configure module?
We need to decide whether to use the configure module or not.
Currently, we're using the module to create 'Configuration' (dict like) instances that load and merge yaml files and to register constructor for yaml tags.
If we decide to use this module, we can delete the 'cli.utils.dict_merge' method because the Configuration class already implements a 'merge' method

aopincar:
DO NOT MERGE YET!!!
In the 'internal' input file we use, there is a use of the old '!env' & '!join' tags.
With this patch we can't use them, so don't while we make use of the 'internal' settings file.

Resolves #45


cli/utils.py
```diff
@@ -62,8 +64,7 @@ def validate_settings_dir(settings_dir=None):
     :raise: IRFileNotFoundException: when the path to the settings dir doesn't
             exist
     """
-    settings_dir = settings_dir or os.environ.get(
-        utils.INFRARED_DIR_ENV_VAR)
+    settings_dir = settings_dir or os.environ.get(INFRARED_DIR_ENV_VAR)
```
yfried-redhat: This method should be part of conf module, and not utils

cli/yamls.py
```diff
@@ -15,95 +13,6 @@
 
 LOG = logger.LOG
```
yfried-redhat: So, This removes the metaclass, and this module no longer needs to be imported on init, does it?
Also, consider renaming this module, or merging it with utils

gamado:

rhosqeauto:

yfried-redhat:
retest this please

gamado:
retest this please

gamado:
retest this please

yfried-redhat:
@aopincar Please add this to our JIRA if it's still relevant. Otherwise, please close this

gamado:
retest

aopincar:
Implemented in #156

Issue #46, opened 2016-02-28
Can 'SpecManager' class be an Abstract class?


aopincar:
The all idea of this class is to create and return parsers which based on spec files - there is no a real need to create and instance of this class for that. Only need to call the 'get_specs' methos with the relevant module name.

tkammer:
To summarize what Ariel is writing here, please make the class a 'Factory' class.

obaranov:
Agree. I will implement that.

yfried-redhat:
Closed by #71

PR #71, opened 2016-03-01
Application object

### LINKED ISSUE #58: Add IRApplication class and encapsulate main logic there
I'd like to add Application class to encapsulate common IR-<name> logic there. This class will have a major method 'run' which will do the following: 
```
 args, settings_files = self.configure()
 settings = self.lookup(settings_files)
 self.dump_settings(settings)
 self.execute(settings)
```

After that we will be able to create main methods for our ir-<name> app's with some thing like that

```
main_provision = IRApplication.create_for('provisioner', CONF).run
main_installer = IRApplication.create_for('installer', CONF).run
main_tester = IRApplication.create_for('tester', CONF).run
```
What do you think on that? Any ideas?

obaranov:
Added application object and other dependent classes to split responsibilities. 
SpecManager is changed to be static

This PR resolves #58 and #46

@tkammer, @yfried-redhat please review.


cli/install.py
```diff
@@ -44,8 +44,7 @@ def get_args(args=None):
     :return args: return loaded arguments from CLI
     """
 
-    spec_manager = conf.SpecManager(CONF)
-    args = spec_manager.parse_args(ENTRY_POINT, args=args)
+    args = conf.SpecManager.parse_args(ENTRY_POINT, CONF, args=args)
```
yfried-redhat: This method can be removed completely

cli/provision.py
```diff
@@ -17,71 +19,163 @@
...
+            app_instance = IRApplication(app_name, setting_dir, args)
+
+        else:
+            raise Exception(
+                "Application is not supported: '{}'".format(app_name))
+        return app_instance
+
+    @classmethod
+    def configure(cls, args):
+        cls._configure_log(args)
```
yfried-redhat: I might be missing something but why do we need _configure_logs method?

cli/provision.py
```diff
@@ -17,71 +19,163 @@
...
+
+    @classmethod
+    def _configure_log(cls, args):
+        """
+        Performs the logging module configuration.
+        :return:
+        """
+        from logging import WARNING, INFO, DEBUG
```
yfried-redhat: Please explain why local import is required (in comment)

cli/provision.py
```diff
@@ -17,71 +19,163 @@
...
+    for the application
+    """
 
-    # playbook execution stage
-    if not args['dry-run']:
-        vars(args)['settings'] = yaml.load(yaml.safe_dump(
-            cli.yamls.Lookup.settings, default_flow_style=False))
-        if not args['cleanup']:
-            vars(args)['provision'] = True
+    def __init__(self, mame, args, settings_dir):
```
yfried-redhat: s/`mame`/`name`/

cli/provision.py
obaranov: I took it form existing code. Will move to the top of module.

cli/provision.py
```diff
@@ -17,71 +19,163 @@
...
+        args = conf.SpecManager.parse_args(app_name, config)
+        setting_dir = utils.validate_settings_dir(
+            CONF.get('defaults', 'settings'))
+        cls.configure(args)
+
+        if app_name in ["provisioner", ]:
+            app_instance = IRApplication(app_name, setting_dir, args)
+
+        else:
+            raise Exception(
```
yfried-redhat: We don't raise general Exceptions. Please use/create a specific IRException

cli/install.py
obaranov: This method also encloses the ENTRY_POINT and CONF module variables. I would prefer to leave method as is. Otherwise we need to add additional imports in all the tests were this method is used.

cli/provision.py
```diff
@@ -17,71 +19,163 @@
...
-    output = yaml.safe_dump(cli.yamls.Lookup.settings,
-                            default_flow_style=False)
-
-    if args['output-file']:
-        with open(args['output-file'], 'w') as output_file:
-            output_file.write(output)
-    else:
-        print output
+    @classmethod
+    def create(cls, app_name, config):
```
yfried-redhat: Do we need to pass `config` to func? `CONF` is shared among all modules, isn't it?

obaranov: I prefer this approach to define all the class dependencies in constructor. Doing that we are independent from the module and it changes. So in future we might change the way we receive configuration. This change won't affect this class.

cli/install.py
yfried-redhat: Then please add a todo "remove online method"

cli/provision.py
```diff
@@ -17,71 +19,163 @@
...
+        self.settings_dir = settings_dir
+        self.sub_command = IRSubCommand.create(name, settings_dir, args)
+
+    def run(self):
+        """
+        :return: Runs the application
+        """
+        settings = self.lookup(self.sub_command.get_settings_files())
+        self.dump_settings(settings)
+        self.execute(settings)
```
yfried-redhat: This is not a blocker, but each application should determine its own execute method and subcommands might wish to call a different playbook than default.

obaranov: We should be able to do that by extending IRApplication class and overriding 'run' method. I'll work on that during Installer implementation.

cli/provision.py
obaranov: fixed. method removed.

obaranov:
@yfried-redhat, I've addressed all the review comments and merged with the master.

yfried-redhat:
fix #46 
fix #58

PR #80, opened 2016-03-02
Revert Installer to old OSPD

yfried-redhat:
Keep the cli code, but revert playbooks, settings and roles

New ir-install has moved away from vision.
Bring back old install code, in order to deliver required functionality.


gamado:

yfried-redhat:
ok to test

PR #96, opened 2016-03-09
Add functionality to load args from ini file

obaranov:
* add additional argument and clg type
 * update spec manager

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,18 +76,18 @@ InfraRed input arguments
 ------------------------
 InfraRed accepts the next sources of the input arguments (in priority order):
 
-#. Command line arguments:  ``ir-provision virsh --host=some.host.com --ssh_user=root``
-#. Predefined arguments in ini file. Use the --from-file option to specify ini file:
+1. Command line arguments:  ``ir-provision virsh --host=some.host.com --ssh_user=root``
+2. Predefined arguments in ini file. Use the --from-file option to specify ini file::
+  
+    ir-provision virsh --host=some.host.com --from-file=user.ini
+  
+    cat user.ini
+    [virsh]
+    ssh_user=root
+    ssh_key=mkey.pm
 
-  ```
-  $ ir-provision virsh --host=some.host.com --from-file=user.ini
-  $ cat user.ini
-  [virsh]
-  ssh_user=root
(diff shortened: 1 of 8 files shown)


rhosqeauto:

cli/conf.py
```diff
@@ -53,7 +75,16 @@ def parse_args(cls, module_name, config, args=None):
         :param module_name: the module name: installer|provisioner|tester
         """
         cmd = clg.CommandLine(cls._get_specs(module_name, config))
-        return cmd.parse(args)
+        res_args = cmd.parse(args)
+
+        # override the command specific arguments from file
```
yfried-redhat: We don't want to override CLI from file.

the search order should be:
1. look for item in CLI.
2. If not found in CLI, look in file (if provided)

yfried-redhat:
Please update README accordingly

cli/conf.py
obaranov: This is exactly what def parse_args(...) is doing. 
If argument is defined in the cli then it will not be overridden by file. For example, 
```
# ir-provision virsh --host=myhost --from-file=provision.ini
# cat 
provision.ini
[virsh]
host=earth
```
This will resuts 'host' = myhost. 
I've updated utils.dict_merge function. It will merge keys only if value from the first dict is None.

README.rst
```diff
@@ -71,6 +71,26 @@ The merging priority order is:
...
+  $ cat user.ini
+  [virsh]
+  ssh_user=root
+  ssh_key=mkey.pm
+  ```
+
+#. Environment variables: ``HOST=earth ir-provision virsh --ssh_user=root``
+
+Command line arguments have the highest priority. All the undefined variables will be replaced by that arguments from file or from environment.
+
```
yfried-redhat: please update "Add new Plugins" section with default details as well

settings/provisioner/virsh/virsh.spec
```diff
@@ -8,35 +8,28 @@ subparsers:
                   host:
                       type: str
                       help: Address/FQDN of the BM hypervisor
```
yfried-redhat: We still need required arguments. need to figure out how to solve that

PR #100, opened 2016-03-09
[RHOSINFRA-14] Add environment lookup for spec arguments

obaranov:
Added env lookup for spec variables. Fixed args loading priority.

### DIFF
--- a/cli/conf.py
+++ b/cli/conf.py
@@ -97,7 +97,9 @@ class SpecManager(object):
         if 'from-file' in res_args:
             file_args = res_args['from-file']
             if file_args is not None and res_args['command0'] in file_args:
-                defaults = utils.dict_merge(file_args[res_args['command0']], defaults)
+                defaults = utils.dict_merge(
+                    file_args[res_args['command0']],
+                    defaults)
 
         # replace defaults with cli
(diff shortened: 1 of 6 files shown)


cli/utils.py
```diff
@@ -184,6 +184,16 @@ def normalize_file(file_path):
     return file_path
 
 
+def string_to_list(value, separator=','):
+    """
+    Converts string to list.
+    """
+    if value.startswith('[') and value.endswith(']'):
```
yfried-redhat: this condition was checked before the call

cli/conf.py
```diff
@@ -75,15 +76,34 @@ def parse_args(cls, module_name, config, args=None):
...
+            module_name,
+            res_args['command0'],
+            DEFAULT_INI
+        )
+        defaults = IniFileType(default_file)[res_args['command0']]
+
+        # override defaults with env variables
+        for arg_name, arg_value in res_args.iteritems():
+            upper_arg_name = arg_name.upper()
+            if upper_arg_name in os.environ:
```
yfried-redhat: If value was provided in CLI, we ignore everything:
`if arg_value is not None and ...`

Issue #34, opened 2016-02-25
make topology mandatory


tkammer:
Make the cleanup playbook use the topology file so it will clean only what the provision created

aopincar:
This issue is a little problematic because it conflicts with issue #33.
If this flag will be mandatory, the parser creation process will fail when trying to execute cleanup with only 'hypervisor' flag

rhosqeauto:
test from prod

Issue #33, opened 2016-02-25
--cleanup should be dependent only on --hypervisor flag


tkammer:
As a user, I should not be forced to provide any params other than the --cleanup and --host/hypervisor flag for the cleanup process to start.

tkammer:
Basically, when a "--cleanup" flag is present, the bare minimum flags are needed to run the cleanup process.
For now, it's host details (host + user/key) and topology (for packages) and future more refine cleanup (that is based on the topology used for installation)

PR #102, opened 2016-03-10
[Rhosinfra-23] dynamically create default.ini from spec file

obaranov:
1) Now default values can be defined in the spec files
2) Removed default.ini file
3) Added generate-conf-file option to generate ini file from the defaults values. 

Several comments: 
 - We cannot remove 'from-file' from spec file because clg will raise then the unsupported argument exception. Anyway from-file argument does not affect default values logic anymore.  
 - dafault.ini file is no longer required. All default values will be stored in spec files. These default values can be overriden by env variables and by the ini file. 

@yfried-redhat please review.


cli/conf.py
```diff
@@ -1,5 +1,7 @@
 import ConfigParser
 
+import sys
+
 import clg
 import os
```
yfried-redhat: AFAIK, imports should be grouped:
1. builtin
2. 3rd party
3. local imports

cli/conf.py
```diff
@@ -74,17 +74,35 @@ def parse_args(cls, module_name, config, args=None):
...
+                out_config.add_section(section)
+                for opt, value in defaults.iteritems():
+                    if value is not None:
+                        out_config.set(section, opt, value)
+                with open(file_name, 'w') as configfile:    # save
+                    out_config.write(configfile)
+            except Exception as ex:
+                raise exceptions.IRException(ex.message)
+            finally:
+                sys.exit(0)
```
yfried-redhat: I don't think we should have sys.exit anywhere here.
I suggest setting `dry-run=True` and replacing output file with the rest of the the output.
Alternatively, generate-conf-file can be a new subcommand, with "virsh" as its input.

yfried-redhat: User should be unable to use generate-conf-file with any other virsh argument (as it has no meaning). Except, maybe, for `-o`/`--output`

cli/conf.py
```diff
@@ -105,6 +123,81 @@ def parse_args(cls, module_name, config, args=None):
 
         return res_args
 
+
+class SpecLoader(object):
```
yfried-redhat: Do we need a whole class of "class methods"? Seems like it would be simpler to make this class a different module (spec.py) with simple functions.

cli/conf.py
```diff
@@ -105,6 +123,81 @@ def parse_args(cls, module_name, config, args=None):
...
+    def load(cls, module_name, config):
+        """
+        Loads the spec file as dict.
+        """
+        clg_options = cls._get_specs(module_name, config)
+        global_options, subcommands_options = cls._get_options(clg_options)
+        return clg_options, global_options, subcommands_options
+
+    @classmethod
+    def _get_options(cls, spec_dict, lookup=False, trim=False):
```
yfried-redhat: This need some documentation.
What's the diff between `_get_options` and `__get_options`?

cli/conf.py
```diff
@@ -105,6 +123,81 @@ def parse_args(cls, module_name, config, args=None):
...
+        return global_options, subcommand_options
+
+    @classmethod
+    def __get_groups_options(cls, options, spec_dict, lookup, trim):
+        if 'groups' in spec_dict:
+            for group in spec_dict['groups']:
+                cls.__get_options(options, group, lookup, trim)
+
+    @classmethod
+    def __get_options(cls, options, spec_dict, lookup, trim):
```
yfried-redhat: Need doc

cli/conf.py
```diff
@@ -105,6 +123,81 @@ def parse_args(cls, module_name, config, args=None):
...
+                    cls.lookup_option(opt_name, opt_params)
+
+                # get a parameters copy with all the keys.
+                options[opt_name] = dict(opt_params)
+
+                if trim:
+                    cls.trim_option(opt_name, opt_params)
+
+    @classmethod
+    def lookup_option(cls, opt_name, opt_params):
```
yfried-redhat: ditto

cli/conf.py
```diff
@@ -105,6 +123,81 @@ def parse_args(cls, module_name, config, args=None):
...
+                        param_value.replace(lookup_string, str(opt_params.get(
+                            lookup_key, lookup_string)))
+
+        # insert default value into help.
+        if 'help' in opt_params and 'default' in opt_params:
+            opt_params['help'] += " | Default value: {}".format(
+                opt_params['default'])
+
+    @classmethod
+    def trim_option(cls, opt_name, opt_params):
```
yfried-redhat: ditto

cli/conf.py
```diff
@@ -105,6 +123,81 @@ def parse_args(cls, module_name, config, args=None):
...
+        # collect sub parser global_options
+        if 'subparsers' in spec_dict:
+            for subparser_name, params in spec_dict['subparsers'].iteritems():
+                subcommand_options[subparser_name], _ = cls._get_options(
+                    params, lookup=True, trim=True)
+
+        return global_options, subcommand_options
+
+    @classmethod
+    def __get_groups_options(cls, options, spec_dict, lookup, trim):
```
yfried-redhat: Need doc

cli/conf.py
yfried-redhat: Guessing: "removes clg attributes (default and required) from clg argument."
If so, please rename "opt_name" and "opt_params". If not, please provide a proper docstring

cli/utils.py
```diff
@@ -49,9 +46,9 @@ def none_resolver(first, second, key):
...
         # tyr to merge lists first
         if isinstance(first[key], list):
-            if isinstance(first[key], list):
+            if isinstance(second[key], list):
                 first[key].extend(second[key])
-            else:
+            elif second[key] is not None:
                 first[key].append(first[key])
 
         if key not in first or first[key] is None:
```
yfried-redhat: Comments about dict_merge:

- dict merge strategy is unclear. need more doc. testing would be helpful as well.
- let's avoid merging lists. values should be overwritten
- the recursive call to dict_merge doesn't use the conflict resovler
- Unless we have a need for different conflict resolvers, I believe that we can allways use the default one.
- If I understand correctly, first[key] is chosen unless it is `None`, in which case, second[key] is chosen. I suggest to always use use second[key] unless:
  - key not in second
  - both are dicts and then invoke recursion.

aopincar:
Hi Oleksii,

Regards ticket #33 in Jira:
I have a unittest ready for dicts merging:  [https://github.com/aopincar/InfraRed/tree/RHOSINFRA-33](url)

About the dict merge, ticket #34 in Jira, I'll try to push a solution by the morning.

cli/conf.py
obaranov: @yfried-redhat can we move that comment to the separate task? This requires major logic change. I'd prefer to implement that as a separate feature.

obaranov:
Review comments addressed: 
1) Spec related code was moved to the separate module.
2) Added method's documentation
3) Replaced conflict resolver for the dict_merger method. None-resolver is still required to merger cli, env, anc ini arguments. List merge is also required for the cli arguments like --extra-vars with action=append. 
4) Added unit tests for dict_merge (including @aopincar test)

gamado:
retest

PR #105, opened 2016-03-11
[Rhosinfra 25] Define required arguments

obaranov:
Dependent on PR: https://github.com/rhosqeauto/InfraRed/pull/102. 

Added 'required' parameter and 'required_only' for spec options. The 'required_only' can list the options that are only required with current option is set. For example for 'cleanup' option with don't need all the required options to be set. 

@yfried-redhat please review.


yfried-redhat:
requires_only is a good addition. however, it is not in clg doc. please add to our doc.
Please add unit-tests ("required" and "requires_only").

README.rst
```diff
@@ -117,15 +119,6 @@ There are two steps that should be done when adding a new plugin to InfraRed:
...
-2. Creating a default spec file (default.ini). 
-    This file should contain the default values for the command line arguments. All the default values should go under the name section names as a new plugin. Example::
-      
-      [virsh]
-      topology=all-in-one.yml
-      network=default.yml
-      ssh-key=~/.ssh/id_rsa
-      ssh-user=root
-
 3. Creating settings files.
```
aopincar: This one should be updated to be the second step now (and not the third).
RST supports '#' in enumerated lists for auto-enumeration. Please us it.

README.rst
```diff
@@ -89,6 +89,8 @@ InfraRed accepts the next sources of the input arguments (in priority order):
 
 3. Environment variables: ``HOST=earth ir-provision virsh --ssh_user=root``
 
+.. note:: The simple ini file with the default values can be generated with: ``ir-povision virsh --generate-conf-file=virsh.ini``
```
aopincar: Please mention the the ini file is based on defaults values from the spec file.

cli/conf.py
```diff
@@ -1,13 +1,10 @@
-import ConfigParser
-
 import clg
+import ConfigParser
```
aopincar: import order:
1. standard library imports
2. related third party imports
3. local application/library specific imports
(https://www.python.org/dev/peps/pep-0008/#imports)

cli/conf.py
```diff
@@ -38,102 +35,6 @@ def load_config_file():
         "Please set it in one of the following paths:\n")
 
 
-def IniFileType(value):
```
aopincar: Function names should be lowercase, with words separated by underscores.
(https://www.python.org/dev/peps/pep-0008/#function-names)

settings/provisioner/virsh/virsh.spec
```diff
@@ -8,28 +8,39 @@ subparsers:
                   host:
                       type: str
                       help: Address/FQDN of the BM hypervisor
+                      required: yes
                   ssh-user:
                       type: str
                       help: User to SSH to the host with
+                      default: root
+                      required: yes
```
aopincar: What's the point of using 'required' if 'default' is defined?
(Relevant for all options with 'required' & 'default' in this file)

settings/provisioner/virsh/virsh.spec
```diff
@@ -8,28 +8,39 @@ subparsers:
                   host:
                       type: str
                       help: Address/FQDN of the BM hypervisor
+                      required: yes
                   ssh-user:
                       type: str
                       help: User to SSH to the host with
+                      default: root
```
aopincar: Please add '(default: __DEFAULT__)' to the option's help wherever 'default' is defined, so the default value will be printed in the help screen.

settings/provisioner/virsh/virsh.spec
```diff
@@ -38,6 +49,7 @@ subparsers:
                   cleanup:
                       action: store_true
                       help: Clean given system instead of provisioning a new one
+                      requires_only: [host, ssh-user,  ssh-key, topology]
```
aopincar: +1
Haven't seen the implementation yet, but liked the idea!
Please document it in the project's README file.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+
+SPEC_EXTENSION = '.spec'
+LOOKUP_DICT = {
+    '__DEFAULT__': "default"
+}
+
+TRIM_PARAMS = ['default', 'required', 'requires_only']
+
+
+def IniFileType(value):
```
aopincar: Function names should be lowercase, with words separated by underscores.
(https://www.python.org/dev/peps/pep-0008/#function-names)

cli/spec.py
```diff
@@ -0,0 +1,244 @@
+import clg
```
aopincar: import order:
1. standard library imports
2. related third party imports
3. local application/library specific imports
(https://www.python.org/dev/peps/pep-0008/#imports)

cli/conf.py
```diff
@@ -38,102 +35,6 @@ def load_config_file():
...
-            for filename in [f for f in filenames
-                             if f.endswith(cls.SPEC_EXTENSION)]:
-                res.append(os.path.join(dirpath, filename))
-
-        return res
-
-
 config = load_config_file()
 
 # update clg types
```
aopincar: It may be more logical to move this to 'spec.py' where the new type is defined...

obaranov: I want to keep that in the config module. The intent is to have all the configuration logic in one place.

settings/provisioner/virsh/virsh.spec
obaranov: This will be added automatically. Check these line: https://github.com/obaranov/InfraRed/blob/RHOSINFRA-25/cli/spec.py#L198-L200

settings/provisioner/virsh/virsh.spec
obaranov: I've added that to have complete picture regarding all the spec arguments. 
Additionally, the default value can be overridden by env or ini file. 'Required' in will raise en exception when provided value is empty.

gamado:
retest

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+TRIM_PARAMS = ['default', 'required', 'requires_only']
+
+
+def IniFileType(value):
+    """
+    The custom type for clg spec
+    :param value: the argument value.
+    :return: dict based on a provided ini file.
+    """
+    _config = ConfigParser.ConfigParser()
```
aopincar: No reason to make '_config" non-public by adding the leading underscore.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    _config.read(value)
```
aopincar: Consider using 'readfp' instead of 'read'

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+
+    d = dict(_config._sections)
```
aopincar: Please use more meaningful names for variables, and try to eliminate call directly to non-public variables, especially when you have methods for that.

Make the code more readability (for example):
res_dict = {}
for section in  config.sections():
    res_dict[section] = {} 
    for option,value in config.items(section):
        res_dict[section][option] = value
...
return res_dict

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+
+SPEC_EXTENSION = '.spec'
+LOOKUP_DICT = {
+    '__DEFAULT__': "default"
+}
+
+TRIM_PARAMS = ['default', 'required', 'requires_only']
+
+
+def clg_ini_file(value):
```
aopincar: Function's name should describes what it does. In this case, this is function isn't related to 'clg'.
I suggest something like "ini_to_dict" "cfg_file_to_dict".

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+            out_config.add_section(section)
+            for opt, value in defaults.iteritems():
+                if value is not None:
+                    out_config.set(section, opt, value)
+            with open(file_name, 'w') as configfile:    # save
+                out_config.write(configfile)
+        except Exception as ex:
+            raise exceptions.IRException(ex.message)
+        finally:
+            sys.exit(0)
```
aopincar: Two things in here:
1. Try to eliminate of calling sys.exit (especially outside the main method)
2. What will happen when exception is raised?!?

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+
+def _get_specs(module_name, config):
+    """
+    Gets specs files as a dict from settings/<module_name> folder.
+    :param module_name: the module name: installer|provisioner|tester
+    """
+    res = {}
+    for spec_file in _get_all_specs(config, subfolder=module_name):
+        spec = yaml.load(
+            open(spec_file),
```
aopincar: Always make sure to close open files objects!
Any python Implementation has its own garbage collecting mechanism, some of them close open file automatically when there are no references to those file objects and some leave responsibility to the OS.
Using context manager (with statement) here will be much better!

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+            Loader=yamlordereddictloader.Loader)
+        utils.dict_merge(res, spec)
+    return res
+
+
+def _get_all_specs(config, subfolder=None):
+    """
+    Gets all the list of all spec files.
```
aopincar: Please add params to the method's docstring

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    res = {}
+    for spec_file in _get_all_specs(config, subfolder=module_name):
+        spec = yaml.load(
+            open(spec_file),
+            Loader=yamlordereddictloader.Loader)
+        utils.dict_merge(res, spec)
+    return res
+
+
+def _get_all_specs(config, subfolder=None):
```
aopincar: You can make this method more atomic by having only one parameter 'root_dir'. It'll be the method caller responsibility to pass the right path. In the method itself, you'll just have to verify that the path exists.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    Gets all the list of all spec files.
+    """
+    root_dir = utils.validate_settings_dir(
+        config.get('defaults', 'settings'))
+    if subfolder:
+        root_dir = os.path.join(root_dir, subfolder)
+
+    res = []
```
aopincar: How about this (which can save the need in iterate the list one more time) ?
```python
res_list = []
for root, _, files in os.walk(root_dir):
    res.extend([os.path.join(root, a_file) for a_file in files if a_file.endswith(SPEC_EXTENSION)])

return res_list
```

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    :param opt_name: the option name to process
+    :param opt_params: the dictionary of the option parameters
+        (help, type, default, etc)
+    """
+    for opt_param_name, param_value in opt_params.iteritems():
+        if opt_param_name in TRIM_PARAMS:
+            opt_params.pop(opt_param_name, None)
+
+
+def _get_specs(module_name, config):
```
aopincar: Please consider combining this method with 'get_all_specs'.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+        if 'requires_only' in arg_param:
+            only_args.extend(arg_param['requires_only'])
+    if len(set(unset_args).intersection(set(only_args))) > 0:
+        raise exceptions.IRConfigurationException(
+                "Required input arguments {} are not set!".format(unset_args))
+
+    return res_args
+
+# ***************************************************************
+# private methods
```
aopincar: There isn't really 'private' in python.
Why do you wanna them to be private? If the all purpose of the methods below is to be helper methods, consider making them non-public with only one leading underscore.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+        dict of options for subcommands
+    """
+    clg_options = _get_specs(module_name, config)
+    global_options, subcommands_options = __process_options(clg_options)
+    return clg_options, global_options, subcommands_options
+
+
+def __process_options(spec_dict, lookup=False, trim=False):
+    """
+    Goes through all the spec options modifies them by removing
```
aopincar: please add params to method's docstring.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    # collect sub parser global_options
+    if 'subparsers' in spec_dict:
+        for subparser_name, params in spec_dict['subparsers'].iteritems():
+            subcommand_options[subparser_name], _ = __process_options(
+                params, lookup=True, trim=True)
+
+    return global_options, subcommand_options
+
+
+def __get_groups_options(options, spec_dict, lookup, trim):
```
aopincar: Usually functions with name that starts with 'get' supposed to return some calculated value.
Please add params to method's docstring.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    Gets the dict of options nested within the spec groups
+    """
+    if 'groups' in spec_dict:
+        for group in spec_dict['groups']:
+            __get_flat_options(options, group, lookup, trim)
+
+
+def __get_flat_options(options, spec_dict, lookup, trim):
```
aopincar: Usually functions with name that starts with 'get' supposed to return some calculated value.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    Gest the dict of options listed in the spec of group.
```
aopincar: Please add params to method's docstring.

cli/spec.py
```diff
@@ -0,0 +1,244 @@
...
+        for subparser_name, params in spec_dict['subparsers'].iteritems():
+            subcommand_options[subparser_name], _ = __process_options(
+                params, lookup=True, trim=True)
+
+    return global_options, subcommand_options
+
+
+def __get_groups_options(options, spec_dict, lookup, trim):
+    """
+    Gets the dict of options nested within the spec groups
```
aopincar: Please add params to method's docstring.

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    elif len(unset_args) > 0:
+        raise exceptions.IRConfigurationException(
+            "Required input arguments {} are not set!".format(
+                unset_args))
+    return res_args
+
+
+def _get_command_line_args(args, config, module_name):
+    """
+    Gets the arguments values from the command line.
```
aopincar: Please add more details about params and return values

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    clg_options, global_opts, subcommand_opts = \
```
aopincar: Where's the use in 'global_opts'? If it's unused, it probably shouldn't been returned.

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+                "Required only input arguments {} are not set!".format(
+                    intersection))
+    elif len(unset_args) > 0:
+        raise exceptions.IRConfigurationException(
+            "Required input arguments {} are not set!".format(
+                unset_args))
+    return res_args
+
+
+def _get_command_line_args(args, config, module_name):
```
aopincar: Please consider rename this function to something more meaningful because it returns cmd line args ('res_args') and subcommand_opts also, or even think of split this function into two.

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    Loads the spec file as dict.
+
+    Returns tuple:
+        dict of options and groups for clg
+        dict of options without subcommands
+        dict of options for subcommands
+    """
+    clg_options = _get_specs(module_name, config)
+    global_options, subcommands_options = _process_options(clg_options)
```
aopincar: What about 'parser_options' & 'subparser_options'?

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    Gets the arguments values from the command line.
+    """
+    clg_options, global_opts, subcommand_opts = \
+        _load_spec_as_dict(module_name, config)
+    cmd = clg.CommandLine(clg_options)
+    res_args = vars(cmd.parse(args))
+    return res_args, subcommand_opts
+
+
+def _load_spec_as_dict(module_name, config):
```
aopincar: This function is not necessarily needed and a bit confusing also, you can call the inside functions from the outside directly

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    # collect sub parser global_options
+    if 'subparsers' in spec_dict:
+        for subparser_name, params in spec_dict['subparsers'].iteritems():
+            subcommand_options[subparser_name], _ = _process_options(
+                params, lookup=True, trim=True)
+
+    return global_options, subcommand_options
+
+
+def _process_groups_options(options, spec_dict, lookup, trim):
```
aopincar: Please consider more meaningful name (ex: 'get_parser_group_options')
Also, create and return dict object inside this function instead of receiving it from the outside.

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    :param lookup: specifies whether the options lookup is required.
+    :param trim: specifies whether the certain option
+        parameters should be removed
+    """
+    if 'groups' in spec_dict:
+        for group in spec_dict['groups']:
+            _process_flat_options(options, group, lookup, trim)
+
+
+def _process_flat_options(options, spec_dict, lookup, trim):
```
aopincar: Same as the comments for the function above:
Please consider more meaningful name (ex: 'get_parser_options')
Also, create and return dict object inside this function instead of receiving it ('options') from the outside.

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+        dict of options and groups for clg
+        dict of options without subcommands
+        dict of options for subcommands
+    """
+    clg_options = _get_specs(module_name, config)
+    global_options, subcommands_options = _process_options(clg_options)
+    return clg_options, global_options, subcommands_options
+
+
+def _process_options(spec_dict, lookup=False, trim=False):
```
aopincar: Please consider more meaningful name (ex: 'get_command_and_subcommand_options')

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+                _lookup_option(opt_name, opt_params)
+
+            # get a parameters copy with all the keys.
+            options[opt_name] = dict(opt_params)
+
+            if trim:
+                _trim_option(opt_name, opt_params)
+
+
+def _lookup_option(opt_name, opt_params):
```
aopincar: Where is the use in 'opt_name'?

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+                    param_value.replace(lookup_string, str(opt_params.get(
+                        lookup_key, lookup_string)))
+
+    # insert default value into help.
+    if 'help' in opt_params and 'default' in opt_params:
+        opt_params['help'] += " | Default value: {}".format(
+            opt_params['default'])
+
+
+def _trim_option(opt_name, opt_params):
```
aopincar: Where is the use in 'opt_name'?

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    # first check __*__ pattern
+    for opt_param_name, param_value in opt_params.iteritems():
+        for lookup_string, lookup_key in LOOKUP_DICT.iteritems():
+            if lookup_string in str(param_value):
+                opt_params[opt_param_name] = \
+                    param_value.replace(lookup_string, str(opt_params.get(
+                        lookup_key, lookup_string)))
+
+    # insert default value into help.
+    if 'help' in opt_params and 'default' in opt_params:
```
aopincar: Please remember that we're using the 'clg' module, user's may add '__DEFAULT__' in option's help. If a user already provided the '__DEFAULT__' in help, there is not need to add this addition.

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+def _trim_option(opt_name, opt_params):
+    """
+    Removes the defined option parameters.
+    :param opt_name: the option name to process
+    :param opt_params: the dictionary of the option parameters
+        (help, type, default, etc)
+    """
+    for opt_param_name, param_value in opt_params.iteritems():
```
aopincar:
```python
for trim_param in TRIM_PARAMS:
    opt_params.pop(trim_param, None)
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    """
+    Looks for all the specs for specified module
+    and parses the commandline input arguments accordingly.
+
+    :param module_name: the module name: installer|provisioner|tester
+    :param config: the infrared configuration file
+    :param args: additional arguments to pass to the clg.
+    """
+
+    res_args, subcommand_opts = _get_command_line_args(args, config,
```
aopincar: It's a bit confusing for me, try to split it into to steps:
1. get clg_opts
2. get subcommand_opt

You can try something like this:
```python
clg_options = _parse_command_line(...)
subcommand_options = _get_subcommand_options(...)
```
 Or maybe even to separate subcommand options from command options in clg_options object
```python
print clg_options
{'command_options': {...},
'subcommand_options': {...}
}
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+                                                       module_name)
+
+    # get the default values
+    command_args = subcommand_opts[res_args['command0']]
+    defaults = {k: v.get('default', None)
```
aopincar: There's no need to hold options with no default value
```python
defaults = {option: attributes['default'] for option, attributes in command_args.iteritems() if 'default' in attributes}
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+                for k, v in command_args.iteritems()}
+
+    # generate config file if required
+    if res_args.get('generate-conf-file', None):
```
aopincar: I see you've removed the call to sys.exit, great!
Do we need to continue with the flow after this if if the user provide the '--generate-conf-file' flag?
If not, you need to break the flow in one of two ways:
1. 'return' in the end of the if block (@tkammer will not like it...)
2. insert the rest of the code into an 'else' block

cli/utils.py
```diff
@@ -195,6 +201,7 @@ def string_to_list(value, separator=',', append_to_list=True):
             value = [value, ]
     return value
```
aopincar: This methods works with dictionaries, which are mutable objects. There is no need to return the object you're passing into the method, someone may think this method return a new dict object...

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+        upper_arg_name = arg_name.upper()
+        if arg_value is None and upper_arg_name in os.environ:
+            defaults[arg_name] = os.getenv(upper_arg_name)
+
+    # override defaults with the ini file args
+    if 'from-file' in res_args:
+        file_args = res_args['from-file']
+        if file_args is not None and res_args['command0'] in file_args:
+            defaults = utils.dict_merge(
+                file_args[res_args['command0']],
```
aopincar: Please take a look on the comment on dict_merge function. (about the return)

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+
+    # replace defaults with cli
+    utils.dict_merge(res_args, defaults,
+                     conflict_resolver=utils.ConflictResolver.none_resolver)
+
+    # check if we have all the required arguments defined at the end of merge
+    unset_args = []
+    only_args = []
+    for arg_name, arg_param in command_args.iteritems():
+        if 'required' in arg_param and \
```
aopincar:
```python
if 'required' in arg_param and \
    arg_param['required'] and \
```
<==>
```python
if arg_param.get('required', None) \
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+                arg_param['required'] and \
+                res_args.get(arg_name, None) is None:
+            unset_args.append(arg_name)
+        if 'requires_only' in arg_param:
+            only_args.extend(arg_param['requires_only'])
+    if len(only_args) > 0:
```
aopincar:
```python
if len(only_args) > 0:
```
<==>
```python
if only_args:
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+        intersection = set(unset_args).intersection(set(only_args))
+        if len(intersection) > 0:
+            raise exceptions.IRConfigurationException(
+                "Required only input arguments {} are not set!".format(
+                    intersection))
+    elif len(unset_args) > 0:
```
aopincar:
```python
elif len(unset_args) > 0:
```
<==>
```python
elif unset_args:
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+    for arg_name, arg_param in command_args.iteritems():
+        if 'required' in arg_param and \
+                arg_param['required'] and \
+                res_args.get(arg_name, None) is None:
+            unset_args.append(arg_name)
+        if 'requires_only' in arg_param:
+            only_args.extend(arg_param['requires_only'])
+    if len(only_args) > 0:
+        intersection = set(unset_args).intersection(set(only_args))
+        if len(intersection) > 0:
```
aopincar:
```python
if len(intersection) > 0:
```
<==>
```python
if intersection:
```

cli/spec.py
```diff
@@ -0,0 +1,260 @@
...
+        for option, value in config.items(section):
+            value = utils.string_to_list(value, append_to_list=False)
+            res_dict[section][option] = value
+
+        res_dict[section].pop('__name__', None)
+
+    return res_dict
+
+
+def parse_args(module_name, config, args=None):
```
aopincar: This function is very long... Try to move the code blocks inside into own functions with meaningful names so it will be easier to read and understand

cli/utils.py
obaranov: This method converts string to the list. So the 'value' is a string and immutable. I will keep return there.

aopincar: Sorry, wrong line... my comment was for the 'dict_merge' method!

cli/utils.py
```diff
@@ -76,7 +82,6 @@ def dict_merge(first, second, path=None,
                 conflict_resolver(first, second, key)
         else:
             first[key] = second[key]
-    return first
```
aopincar: Please take a look on the comment a gave on the return value in 'string_to_list' method by mistake.

PR #124, opened 2016-03-21
 RHOSINFRA-52 Add 'openstack' provisioner

bregman-arie:
Added OpenStack provisioner.

It is used for creating and/or removing resources on OpenStack clouds.

Includes:

* provision networks
    - Allows to specify if network is shared
    - Allows to specify timeout for network
* provisiom subnets
    - Allows to specify cidr, dns servers, allocation pool
      and whether DHCP should be enabled
* provision routers
    - Allows to attach network interfaces
    - Allows to specify default gateway
* provision instances
    - Allows to specify config_drive, nics, image and flavor
* provision flavors
    - Allows to specify ram, disk and vcpus
* provision floating IPs
    - Allows to reuse existing floating IPs
    - Allows to specify fixed IP address

The style is based on:
* 4 spaces
* module variables are sorted alphabetically
* blank line after each task

### DIFF
--- a/playbooks/provisioner/openstack/cleanup.yaml
+++ b/playbooks/provisioner/openstack/cleanup.yaml
@@ -1,0 +1,6 @@
+---
+- name: Cleanup OpenStack resources
+  hosts: localhost
+  roles:
+      - { role: openstack/nova, action: cleanup }
+      - { role: openstack/neutron, action: cleanup }
--- a/playbooks/provisioner/openstack/main.yaml
+++ b/playbooks/provisioner/openstack/main.yaml
@@ -1,0 +1,6 @@
+---
+- name: Provision OpenStack resources
+  hosts: localhost
+  roles:
+      - { role: openstack/neutron, action: setup }
+      - { role: openstack/nova, action: setup }
--- a/plugins/filters/add_prefix.py
+++ b/plugins/filters/add_prefix.py
@@ -1,0 +1,19 @@
+from ansible import utils
+
+
+def add_prefix(value, prefix=''):
+    """Add string prefix to item.
+
+    {{ ['hanukkah', 'easter', 'pessah'] | map('add_prefix', 'happy ' }}
+    -> ['happy hanukkah', 'happy easter', 'happy pessah']
+    """
+
+    return ''.join((prefix, value))
+
+
+class FilterModule(object):
+
+    def filters(self):
+        return {
+            'add_prefix': add_prefix,
+        }
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ ansible>=2.0.0
 clg>=2.0.0
 configure>=0.5
 colorlog>=2.6.1
+python-novaclient<3
 PyYAML>=3.11
 shade>=1.7.0
 yamlordereddictloader>=0.1.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,7 @@
 ansible>=2.0.0
-python-novaclient<3.0.0
-python-neutronclient>=4.0.0
-PyYAML>=3.11
+clg>=2.0.0
 configure>=0.5
 colorlog>=2.6.1
-clg>=2.0.0
+PyYAML>=3.11
+shade>=1.7.0
 yamlordereddictloader>=0.1.1
--- a/roles/openstack/neutron/defaults/main.yml
+++ b/roles/openstack/neutron/defaults/main.yml
@@ -16,3 +16,7 @@ dns_servers:
 # Router defaults
 
 router_name: router1
+
+# Security group defaults
+
+security_group_name: default
--- a/roles/openstack/neutron/defaults/main.yml
+++ b/roles/openstack/neutron/defaults/main.yml
@@ -1,0 +1,18 @@
+# General defaults
+
+prefix: "{{ provisioner.prefix | default(random.uuid) }}"
+
+# Network defaults
+
+timeout_network: 180
+
+# Subnet defaults
+
+enable_dhcp: True
+dns_servers:
+    - 208.67.222.222
+    - 8.8.8.8
+
+# Router defaults
+
+router_name: router1
--- a/roles/openstack/neutron/tasks/cleanup.yml
+++ b/roles/openstack/neutron/tasks/cleanup.yml
@@ -1,0 +1,29 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  os_router:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      name: "{{ prefix }}{{ item.value.name | default(router_name) }}"
+      state: absent
+  with_dict: "{{ provisioner.neutron.routers }}"
+  when: provisioner.neutron.routers is defined
+  ignore_errors: yes
+
+- name: Cleanup subnets
+  os_subnet:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      name: "{{ prefix }}{{ item.value.name }}"
+      state: absent
+  with_dict: "{{ provisioner.neutron.subnets }}"
+  when: provisioner.neutron.subnets is defined
+  ignore_errors: yes
+
+- name: Cleanup networks
+  os_network:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      name: "{{ prefix }}{{ item.value.name }}"
+      state: absent
+  with_dict: "{{ provisioner.neutron.networks }}"
+  when: provisioner.neutron.networks is defined
+  ignore_errors: yes
--- a/roles/openstack/neutron/tasks/cleanup.yml
+++ b/roles/openstack/neutron/tasks/cleanup.yml
@@ -8,6 +8,7 @@
       state: absent
   with_dict: "{{ provisioner.neutron.routers }}"
   when: provisioner.neutron.routers is defined
+  register: routers
   ignore_errors: yes
 
 - name: Cleanup subnets
@@ -17,6 +18,7 @@
       state: absent
   with_dict: "{{ provisioner.neutron.subnets }}"
   when: provisioner.neutron.subnets is defined
+  register: subnets
   ignore_errors: yes
 
 - name: Cleanup networks
@@ -26,4 +28,19 @@
       state: absent
   with_dict: "{{ provisioner.neutron.networks }}"
   when: provisioner.neutron.networks is defined
+  register: networks
   ignore_errors: yes
+
+#TODO(abregman): Enable 'security groups' task when ansible 2.1 is out
+#- name: Cleanup security groups
+#  os_security_group:
+#      cloud: "{{ provisioner.cloud | default(omit) }}" 
+#      name: "{{ prefix }}{{ item.value.name | default(security_group_name) }}" 
+#      state: absent
+#  register: security_groups
+#  with_dict: "{{ provisioner.neutron.security_groups }}" 
+#  when: provisioner.neutron.security_groups is defined
+
+- name: Fail run if one of neutron cleanups tasks failed/skipped
+  fail: msg="The run failed because one of neutron cleanup tasks failed/skipped"
+  when: (routers|skipped or subnets|skipped or networks|skipped)
--- a/roles/openstack/neutron/tasks/main.yml
+++ b/roles/openstack/neutron/tasks/main.yml
@@ -1,0 +1,1 @@
+- include: "{{ action }}.yml"
--- a/roles/openstack/neutron/tasks/setup.yml
+++ b/roles/openstack/neutron/tasks/setup.yml
@@ -52,3 +52,26 @@
   register: routers
   with_dict: "{{ provisioner.neutron.routers }}"
   when: provisioner.neutron.routers is defined
+
+#TODO(abregman): Enable 'security groups' tasks when ansible 2.1 is out
+#- name: Create security groups
+#  os_security_group:
+#      cloud: "{{ provisioner.cloud | default(omit) }}"
+#      name: "{{ prefix }}{{ item.value.name | default(security_group_name) }}"
+#      state: present
+#  register: security_groups
+#  with_dict: "{{ provisioner.neutron.security_groups }}"
+#  when: provisioner.neutron.security_groups is defined
+
+#- name: Create security group rules
+#  os_security_group_rule:
+#      cloud: "{{ provisioner.cloud | default(omit) }}"
+#      security_group: "{{ prefix }}{{ item.value.name | default(security_group_name) }}"
+#      protocol: "{{ item.value.protocol | default(omit) }}"
+#      port_range_min: "{{ item.value.port_range_min | default(omit) }}"
+#      port_range_max: "{{ item.value.port_range_max | default(omit) }}"
+#      remote_ip_prefix: "{{ item.value.remote_ip_prefix | default(omit) }}"
+#      state: present
+#  register: security_group_rules
+#  with_dict: "{{ provisioner.neutron.security_groups.rules }}"
+#  when: provisioner.neutron.security_groups is defined
--- a/roles/openstack/neutron/tasks/setup.yml
+++ b/roles/openstack/neutron/tasks/setup.yml
@@ -1,0 +1,54 @@
+---
+# This file is used to create OpenStack neutron resources
+
+- name: Create networks
+  os_network:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      external: "{{ item.value.external | default(omit) }}"
+      name: "{{ prefix }}{{ item.value.name }}"
+      state: present
+      shared: "{{ item.value.shared | default(omit) }}"
+      timeout: "{{ item.value.timeout | default(timeout_network) }}"
+  register: networks
+  with_dict: "{{ provisioner.neutron.networks }}"
+  when: provisioner.neutron.networks is defined
+
+- name: Register externally accessible networks
+  os_networks_facts:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      filters:
+          router:external: true
+  register: networks_details
+- debug: var=networks_details
+
+- name: Set external network id
+  set_fact:
+      external_network_id: "{{ item.value[0].id }}"
+  with_dict: "{{ networks_details.ansible_facts }}"
+
+- name: Create subnets
+  os_subnet:
+      allocation_pool_start: "{{ item.value.allocation_pool_start }}"
+      allocation_pool_end: "{{ item.value.allocation_pool_end }}"
+      cidr: "{{ item.value.cidr | default(omit) }}"
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      dns_nameservers: "{{ item.value.dns_servers | default(dns_servers) }}"
+      enable_dhcp: "{{ item.value.enable_dhcp | default(enable_dhcp) }}"
+      name: "{{ prefix }}{{ item.value.name }}"
+      network_name: "{{ prefix }}{{ item.value.network_name }}"
+      state: present
+  register: subnets
+  with_dict: "{{ provisioner.neutron.subnets }}"
+  when: provisioner.neutron.subnets is defined
+
+- name: Create routers
+  os_router:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      external_fixed_ips: "{{ item.value.external_fixed_ips | default(omit) }}"
+      interfaces: "{{ item.value.attach_subnets | map('add_prefix', prefix) | join(',') }}"
+      name: "{{ prefix }}{{ item.value.name | default(router_name) }}"
+      network: "{{ external_network_id }}"
+      state: present
+  register: routers
+  with_dict: "{{ provisioner.neutron.routers }}"
+  when: provisioner.neutron.routers is defined
--- a/roles/openstack/nova/defaults/main.yml
+++ b/roles/openstack/nova/defaults/main.yml
@@ -1,0 +1,15 @@
+# General defaults
+
+prefix: "{{ provisioner.prefix | default(random.uuid) }}"
+
+# Instances defaults
+
+auto_ip: no
+config_drive: no
+instance_image: centos-minimal
+wait: yes
+
+# Flavor defaults
+
+flavor_id: 4
+flavor_name: ir-flavor
--- a/roles/openstack/nova/tasks/cleanup.yml
+++ b/roles/openstack/nova/tasks/cleanup.yml
@@ -1,0 +1,45 @@
+---
+# This file is used to cleanup OpenStack nova resources
+
+- name: Register nova servers details
+  os_server_facts:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      server: "{{ prefix }}*"
+  register: servers_details
+- debug: var=servers_details
+
+- name: Set floating IP address
+  set_fact:
+      floating_ip_address: "{{ item.value[0].interface_ip }}"
+  with_dict: servers_details.ansible_facts
+  ignore_errors: yes
+
+- name: Release floating IPs
+  os_floating_ip:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      server: "{{ prefix }}{{ item.value.name }}"
+      floating_ip_address: "{{ floating_ip_address }}"
+      state: absent
+  with_dict: "{{ provisioner.nodes }}"
+  when: provisioner.nodes is defined
+  ignore_errors: yes
+
+- name: Cleanup instances
+  os_server:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      image: "{{ item.value.image | default(instance_image) }}"
+      name: "{{ prefix }}{{ item.value.name }}"
+      state: absent
+  with_dict: "{{ provisioner.nodes }}"
+  when: provisioner.nodes is defined
+  ignore_errors: yes
+
+###todo(abregman): Enable 'Cleanup flavors' when ansible 2.1 is out
+#- name: Cleanup flavors
+#  os_nova_flavor:
+#      cloud: "{{ provisioner.cloud | default(omit) }}"
+#      name: "{{ prefix }}{{ item.value.nane | default(flavor_name) }}"
+#      state: absent
+#  with_dict: "{{ provisioner.flavors }}"
+#  when: provisioner.flavors is defined
+#  ignore_errors: yes
--- a/roles/openstack/nova/tasks/cleanup.yml
+++ b/roles/openstack/nova/tasks/cleanup.yml
@@ -1,40 +1,17 @@
 ---
 # This file is used to cleanup OpenStack nova resources
 
-- name: Register nova servers details
-  os_server_facts:
-      cloud: "{{ provisioner.cloud | default(omit) }}"
-      server: "{{ prefix }}*"
-  register: servers_details
-- debug: var=servers_details
+- include: "cleanup_nodes.yml"
+  with_dict: provisioner.nodes
 
-- name: Set floating IP address
-  set_fact:
-      floating_ip_address: "{{ item.value[0].interface_ip }}"
-  with_dict: servers_details.ansible_facts
-  ignore_errors: yes
-
-- name: Release floating IPs
-  os_floating_ip:
-      cloud: "{{ provisioner.cloud | default(omit) }}"
-      server: "{{ prefix }}{{ item.value.name }}"
-      floating_ip_address: "{{ floating_ip_address }}"
+- name: Remove keypair
+  os_keypair:
+      cloud: "{{ provisioner.cloud | default(omit) }}" 
+      name: "{{ prefix }}{{ provisioner.key.file | basename }}" 
       state: absent
-  with_dict: "{{ provisioner.nodes }}"
-  when: provisioner.nodes is defined
   ignore_errors: yes
 
-- name: Cleanup instances
-  os_server:
-      cloud: "{{ provisioner.cloud | default(omit) }}"
-      image: "{{ item.value.image | default(instance_image) }}"
-      name: "{{ prefix }}{{ item.value.name }}"
-      state: absent
-  with_dict: "{{ provisioner.nodes }}"
-  when: provisioner.nodes is defined
-  ignore_errors: yes
-
-###todo(abregman): Enable 'Cleanup flavors' when ansible 2.1 is out
+#TODO(abregman): Enable 'Cleanup flavors' when ansible 2.1 is out
 #- name: Cleanup flavors
 #  os_nova_flavor:
 #      cloud: "{{ provisioner.cloud | default(omit) }}"
--- a/roles/openstack/nova/tasks/cleanup_nodes.yml
+++ b/roles/openstack/nova/tasks/cleanup_nodes.yml
@@ -1,0 +1,22 @@
+---
+# This file includes tasks for instances and floating IPs cleanup.
+
+# Node type can be: controller, compute, ceph, etc.
+- name: Set fact for node type
+  set_fact:
+      node_type: "{{ item }}"
+
+# These tasks done for each type of node according to 'amount' variable
+- name: Cleanup instances
+  os_server:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      name: "{{ prefix }}{{ node_type.value.name }}-{{ item }}"
+      state: absent
+  with_sequence: "count={{ node_type.value.amount }}"
+  when: provisioner.nodes is defined
+  register: instances
+  ignore_errors: yes
+
+- name: Fail run if one of nova cleanups tasks failed/skipped
+  fail: msg="The run failed because one of nova cleanup tasks failed/skipped"
+  when: instances|skipped
--- a/roles/openstack/nova/tasks/setup.yml
+++ b/roles/openstack/nova/tasks/setup.yml
@@ -1,0 +1,43 @@
+---
+# This file is used to create OpenStack nova resources
+
+#TODO(abregman): Enable 'create_flavors' when ansible 2.1 is out
+#- name: Create flavors
+#  os_nova_flavor:
+#      cloud: "{{ provisioner.cloud | default(omit) }}"
+#      name: "{{ prefix }}{{ item.value.nane | default(flavor_name) }}"
+#      ram: "{{ item.value.ram | default(omit) }}"
+#      disk: "{{ item.value.disk | default(omit) }}"
+#      vcpus: "{{ item.value.vcpus | default(omit) }}"
+#      state: present
+#  register: flavors
+#  with_dict: "{{ provisioner.flavors }}"
+#  when: provisioner.flavors is defined  
+
+- name: Create instances
+  os_server:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      auto_ip: "{{ item.value.auto_ip | default(auto_ip) }}"
+      config_drive: "{{ item.value.config_drive | default(config_drive) }}"
+      flavor: "{{ item.value.flavor | default(flavor_id) }}"
+      image: "{{ provisioner.image }}"
+      name: "{{ prefix }}{{ item.value.name }}"
+      nics: "{{ networks.results | map(attribute='id') | list() |map('add_prefix', 'net-id=') | list() }}"
+      key: "{{ item.value.key | default(omit) }}"
+      key_name: "{{ item.value.key_name | default(omit) }}"
+      state: present
+      wait: "{{ item.value.wait | default(wait) }}"
+  register: nodes
+  with_dict: "{{ provisioner.nodes }}"
+  when: provisioner.nodes is defined
+
+- name: Create floating IPs
+  os_floating_ip:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      network: "{{ external_network_id }}"
+      internal_network: "{{ prefix}}{{ item.value.network.floating_ip_network }}"
+      server: "{{ prefix }}{{ item.value.name }}"
+      state: present
+  register: floating_ips
+  with_dict: "{{ provisioner.nodes }}"
+  when: provisioner.nodes is defined
--- a/roles/openstack/nova/tasks/setup.yml
+++ b/roles/openstack/nova/tasks/setup.yml
@@ -14,30 +14,19 @@
 #  with_dict: "{{ provisioner.flavors }}"
 #  when: provisioner.flavors is defined  
 
-- name: Create instances
-  os_server:
-      cloud: "{{ provisioner.cloud | default(omit) }}"
-      auto_ip: "{{ item.value.auto_ip | default(auto_ip) }}"
-      config_drive: "{{ item.value.config_drive | default(config_drive) }}"
-      flavor: "{{ item.value.flavor | default(flavor_id) }}"
-      image: "{{ provisioner.image }}"
-      name: "{{ prefix }}{{ item.value.name }}"
-      nics: "{{ networks.results | map(attribute='id') | list() |map('add_prefix', 'net-id=') | list() }}"
-      key: "{{ item.value.key | default(omit) }}"
-      key_name: "{{ item.value.key_name | default(omit) }}"
-      state: present
-      wait: "{{ item.value.wait | default(wait) }}"
-  register: nodes
-  with_dict: "{{ provisioner.nodes }}"
-  when: provisioner.nodes is defined
+# This is needed for os_keypair as it uses public key
+- name: Retrieve public key from private key
+  shell: "ssh-keygen -y -f {{ provisioner.key.file }}"
+  register: public_key_content
 
-- name: Create floating IPs
-  os_floating_ip:
+- name: Add keypair
+  os_keypair:
       cloud: "{{ provisioner.cloud | default(omit) }}"
-      network: "{{ external_network_id }}"
-      internal_network: "{{ prefix}}{{ item.value.network.floating_ip_network }}"
-      server: "{{ prefix }}{{ item.value.name }}"
+      name: "{{ prefix }}{{ provisioner.key.file | basename }}"
+      public_key: "{{ public_key_content.stdout }}"
       state: present
-  register: floating_ips
+  when: not provisioner.key.name
+
+# This includes the tasks for instances and floating IPs creation
+- include: "setup_nodes.yml"
   with_dict: "{{ provisioner.nodes }}"
-  when: provisioner.nodes is defined
--- a/roles/openstack/nova/tasks/setup_nodes.yml
+++ b/roles/openstack/nova/tasks/setup_nodes.yml
@@ -1,0 +1,50 @@
+---
+# This file includes tasks for instances and floating IPs creation.
+
+# Node type can be: controller, compute, ceph, etc.
+- name: Set fact for node type
+  set_fact:
+      node_type: "{{ item }}"
+
+# These tasks done for each type of node according to 'amount' variable.
+- name: Create instances
+  os_server:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      auto_ip: "{{ node_type.value.auto_ip | default(auto_ip) }}"
+      config_drive: "{{ node_type.value.config_drive | default(config_drive) }}"
+      flavor: "{{ node_type.value.flavor | default(flavor_id) }}"
+      image: "{{ provisioner.image }}"
+      name: "{{ prefix }}{{ node_type.value.name }}-{{ item }}"
+      nics: "{{ networks.results | map(attribute='id') | list() |map('add_prefix', 'net-id=') | list() }}"
+      key_name: "{{ provisioner.key.name if provisioner.key.name else (prefix + provisioner.key.file | basename) }}"
+      state: present
+      wait: "{{ node_type.value.wait | default(wait) }}"
+  register: nodes
+  with_sequence: "count={{ node_type.value.amount }}"
+  when: provisioner.nodes is defined
+
+- name: Create floating IPs
+  os_floating_ip:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      network: "{{ external_network_id }}"
+      internal_network: "{{ prefix}}{{ node_type.value.network.floating_ip_network }}"
+      server: "{{ prefix }}{{ node_type.value.name }}-{{ item }}"
+      state: present
+  register: floating_ips
+  with_sequence: "count={{ node_type.value.amount }}"
+  when: provisioner.nodes is defined
+
+- name: Gather created servers facts
+  os_server_facts:
+      server: "{{ prefix }}{{ node_type.value.name }}*"
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+  register: servers
+      
+- name: Add created servers to inventory
+  add_host:
+      name: "{{ prefix }}{{ node_type.value.name}}-{{item}}"
+      ansible_ssh_host: "{{ servers.ansible_facts.openstack_servers[0].public_v4 }}"
+      groups: "{{ provisioner.nodes[node_type.value.name].groups | join(',') }}"
+      ansible_ssh_private_key_file: "{{ provisioner.key.file }}"
+      ansible_ssh_user: 'cloud-user'
+  with_sequence: "count={{ node_type.value.amount }}"
--- a/settings/provisioner/openstack/neutron/default.yml
+++ b/settings/provisioner/openstack/neutron/default.yml
@@ -1,0 +1,37 @@
+---
+# This is the most common network topology for OpenStack Testing deployments
+
+networks:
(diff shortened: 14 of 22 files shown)


rhosqeauto:

bregman-arie:
Few points I thought/wondered about while writing this:

* Is there InfraRed style guide? If not, where should it be added?

* I structured the new openstack playbooks in a way I think is the easier for everyone to read + according to Ansible best practices. Please let me know what do you think.

* right now I have defaults per sub role. for example:
  roles/
      openstack/
          network/
              defaults/main.yml

Should we have something more micro/specific like the following?:
  roles/
      openstack/
          network/
              defaults/
                   network.yml
                   subnet.yml
                   router.yml


* roles/openstack/storage - at first I wrote such role but then I realised we probably don't have a use case for it. if you think differently, let me know.

playbooks/provisioner/openstack/main.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Provision OpenStack resources
+  hosts: localhost
+  roles:
+      - openstack/provision/network
```
mpryc: I would much more prefer to stick to OpenStack naming (neutron, nova, glance, cinder)...

playbooks/provisioner/openstack/cleanup.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Cleanup OpenStack resources
+  hosts: localhost
+  roles:
+      - openstack/cleanup/compute
```
mpryc: I would much more prefer to stick to OpenStack naming (neutron, nova, glance, cinder)...

bregman-arie: Sure. Sounds reasonable :+1:

playbooks/provisioner/openstack/main.yaml
bregman-arie: Sure. Sounds reasonable :+1:

roles/openstack/cleanup/compute/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+# Instances defaults
+
+instance_image: centos-minimal
```
mpryc: What is this role for and how it's being used in InfraRed?

roles/openstack/provision/compute/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
+---
```
mpryc: I really like this file.

roles/openstack/cleanup/compute/defaults/main.yml
bregman-arie: Cleanup compute/nova resources

roles/openstack/provision/network/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
...
+external_network: 180
+timeout_network: 180
+shared_network: False
+
+# Subnet defaults
+
+enable_dhcp: True
+dns_servers:
+    - 8.8.8.7
+    - 8.8.8.8
```
mpryc: why ggl dns, maybe first from other provider?

roles/openstack/provision/network/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+      timeout: {{ item.value.timeout | default(timeout_network) }}
+  register: networks
+  with_dict: {{ provisioner.networks }}
+  when: provisioner.networks is defined
+
+- name: Create subnets
+  os_subnet:
+      allocation_pool_start: {{ item.value.allocation_pool_start }}
+      allocation_pool_end: {{ item.value.allocation_pool_end }}
+      cidr: 192.168.0.0/24
```
mpryc: what about ipv6?

bregman-arie: lol. this is actually the wrong value. will change it. thanks.

roles/openstack/provision/network/defaults/main.yml
bregman-arie: hmm simply because it's the most used dns today...
by provider you mean what exactly? anther project/tenant? anther cloud? anther company? :)

mpryc:
RE: micro/specific / "right now I have defaults per sub role. for example"

I really don't see the difference for using one over another, just having one main file to include dir.

What benefits do you see in having such layout?

roles/openstack/provision/network/defaults/main.yml
mpryc: Like opendns: 208.67.222.222

roles/openstack/cleanup/compute/defaults/main.yml
mpryc: Well I asked more how it's ment to use centos-minimal and how it works ;)

bregman-arie: Oh sorry.
The file you specificly asked about is a file of default values.
For creating a server/instance you need a server name and image. Now what happens if user provided name but not an image? we can choose:
1. fail the run
2. use default image. using default image is risky but you might not have this default image in your cloud. I'm not aware of default common image that can be used here. we can still decide option 1 is better.

roles/openstack/provision/network/defaults/main.yml
bregman-arie: cool. I actually wasn't familiar of it. 
so what is exactly the advantage here? I mean why opendns > google dns

yfried-redhat:
@bregman-arie,
About style:
Your style is good.
We don't have guidelines yet. If you feel like drafting them, please add a section to README.
We want to add ansible-lint to the pep8 tests

yfried-redhat:
@bregman-arie I'm good with the general approach of roles, at this time. I'm afraid I'll have to see an actual provisioner including settings to give more details. I foresee some issues with resources if we don't use unique names.
I also wonder about the way to specify the network topology.

tkammer:
@bregman-arie I believe this is a good patch but I do have some general concerns that I would like to raise.
In regards to ansible roles, when creating a role one should ask himself "This machine is: <role>", so something like:
This machine is "nova" or This machine is "Neutron" is acceptable but doing something like "roles/openstack/provison/neutron" falls out of this.
Again, this is my understanding of "roles" while reading the docs as ansible is an orchestrator, so it assigns "roles" to nodes and acts accordingly.

I believe the correct approach should be defining a role such as: roles/openstack/neutron  and under it have the whole structure for that role, example:

Neutron
├── README.md
├── defaults
│   └── main.yml
├── files
├── handlers
│   └── main.yml
├── meta
│   └── main.yml
├── tasks
│   └── main.yml  <-- this file will reference setup.yml and cleanup.yml
│   └── setup.yml
│   └── cleanup.yml
├── templates
└── vars
    └── main.yml

This structure allow us to target a specific node as "Neutron" and deploy everything on it per our need rather than use:

roles:
  - {role: openstack/neutron/setup}
  - {role: openstack/neutron/run}
  - {role: openstack/neutron/cleanup}

Which IMHO doesn't make sense...

If we want to reuse or create the equivalent of "functions", that is either "tasks" or "handlers" rather than "roles".
My 2 cents..

bregman-arie:
@tkammer this is actually great. I actually wasn't sure about the whole cleanup/provision thing also...and I'm totally fine with the structure you proposed. +2. will change it.

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,49 @@
+---
+subparsers:
+    openstack:
+        help: Provision systems using Ansible OpenStack modules
+        groups:
+            - title: image
+              options:
+                  image-name:
+                      type: str
```
yfried-redhat: use "Value" argument instead of "str". see README

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,49 @@
...
+                      help: An image on OpenStack cloud to provision the instance with
+                      required: yes
+            - title: topology
+              options:
+                  network:
+                      type: str
```
yfried-redhat: use "YamlFile" arg type

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,49 @@
...
+                      help: Network
+                      default: default.yml
+                  topology:
+                      type: str
```
yfried-redhat: use "Topology" arg type

settings/provisioner/openstack/openstack.spec
bregman-arie: ack
thanks

settings/provisioner/openstack/network/default.yml
```diff
@@ -0,0 +1,40 @@
...
+            network_name: !lookup provisioner.networks.management.name
+            allocation_pool_start: 192.168.1.10
+            allocation_pool_start: 192.168.1.100
+            cidr: 192.168.1.0/24
+
+    rotuers:
+        router:
+            name: '{{ run.uuid }}-router'
+            attach_subnets:
+                - !lookup provisioner.subnets.external.name
```
obaranov: Do not reference here to another lookup. You can just do !lookup provisioner.networks.external.name
The same is actual for the second item in the list 

Also I think all the lookup should be surrounded with "{{ }}"

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,26 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
```
yfried-redhat: All these arguments are wrong. you need to use `"{{ arg }}"`  and not `{{ arg }}`.

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,26 @@
...
+- name: Cleanup subnets
+  os_subnet:
+      name: {{ item.value.name }}
+      state: absent
+  with_dict: {{ provisioner.subnets }}
+  when: provisioner.subnets is defined
+
+- name: Cleanup networks
+  os_network:
+      name {{ item.value.name }}
```
yfried-redhat: typo

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,49 @@
...
+                      type: Value
+                      help: An image on OpenStack cloud to provision the instance with
+                      required: yes
+            - title: topology
+              options:
+                  neutron:
+                      type: YamlFile
+                      help: Network resources
+                      default: default.yml
+                  topology:
```
yfried-redhat: this will map to `provisioner.topology.topolgy` I suggest to rename this to `nodes`, or, `nova`

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,55 @@
+---
+subparsers:
+    openstack:
+        help: Provision systems using Ansible OpenStack modules
+        groups:
+            - title: cloud
+              options:
+                  cloud:
+                      type: Value
```
tkammer: I assume this needs to point to a cloud file with params no?
I would say you need this as YamlFile type not "value".

yfried-redhat:
About `--cloud`:
Working with shade opens a lot of possibilities re input, lets use them all:

- if cloud/auth aren't supplied shade will lookup OS_* env vars (as provided by keystonerc files for example). This allows all users to work as they are used to. Make sure this is possible and documented
- Help should also explain where `cloud` is pulling data from.
- `cloud` in in a different group (which is great). keep it there
- users might wish to provide creds explicitly. Please add `--auth-auth_url`, `--auth-username`, `--auth-password`, and `--auth-project_name` as optional arguments. Shade allows any of them to override `cloud` data. Auto nesting will nest them in dict under "auth" which you can load to modules' "auth" argument and omit by default. Don't forget proper doc for these as well

settings/provisioner/openstack/openstack.spec
yfried-redhat: This is actually just a string. but It does need more documentation. 
Please make sure the help text explains how to use this argument properly.

settings/provisioner/openstack/neutron/default.yml
```diff
@@ -0,0 +1,38 @@
+---
+# This is the most common network topology for OpenStack Testing deployments
+
+networks:
+    data:
+        name: '{{ random.uuid }}-data'
```
yfried-redhat: I suggest we add the input argumet "prefix" if users wish to avoid the randomizer (ie for cleanup):
Then, we can do:
`'{{ provisioner.prefix | default(random.uuid) }}-data'`

That will try to use "prefix", and if missing, use the random prefix

yfried-redhat: and cleanup will be triggered by `... --cleanup --prefix=fdsts`

settings/provisioner/openstack/neutron/default.yml
```diff
@@ -0,0 +1,38 @@
...
+        cidr: 172.31.0.0/23
+    management:
+        name: "{{ !lookup provisioner.neutron.networks.management.name }}"
+        network_name: "{{ !lookup provisioner.neutron.networks.management.name }}"
+        allocation_pool_start: 192.168.1.10
+        allocation_pool_end: 192.168.1.100
+        cidr: 192.168.1.0/24
+
+routers:
+    router:
```
yfried-redhat: Why do we need `router.routers`?
Seems like a list would be better:

```yaml
routers:
   - name: '{{ prfix }}-router'
     network: "whatever"
   - name: '{{ prfix }}-another-router-just-because-I-want'
     network: "whatever"
```

roles/openstack/nova/tasks/setup.yml
```diff
@@ -0,0 +1,47 @@
...
+- name: Create instances
+  os_server:
+      cloud: "{{ provisioner.cloud }}"
+      auto_ip: "{{ item.value.auto_ip | default(omit) }}"
+      floating_ip_pools: "{{ provisioner.floating_ip_pool | default(omit) }}"
+      config_drive: "{{ item.value.config_drive | default(config_drive) }}"
+      flavor: "{{ item.value.flavor | default(flavor_id) }}"
+      image: "{{ provisioner.image }}"
+      name: "{{ item.value.name | default(instance_name) }}"
+      nics:
```
yfried-redhat: `{% for net in networks.results %} net-id={{ net.id }}{% endfor }`

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,55 @@
...
+                      help: The cloud which the OpenStack modules will operate against
+                      required: yes
+            - title: image
+              options:
+                  image:
+                      type: Value
+                      help: An image id or name, on OpenStack cloud to provision the instance with
```
yfried-redhat: since we have shade available, Please add the shade command to list all images to the help txt.

roles/openstack/nova/tasks/cleanup.yml
```diff
@@ -0,0 +1,24 @@
+---
+# This file is used to cleanup OpenStack nova resources
+
+- name: Release floating IPs
+  os_floating_ip:
+      server: "{{ item.value.server | default(instance_name) }}"
+      state: absent
+  with_dict: "{{ provisioner.floating_ips }}"
+  when: provisioner.floating_ips is defined
```
yfried-redhat: Where can `floating_ips` be defined? seems like we are going to have a leakage here.
Also, Have you verified that "absent" will actually delete the floater? I seem to recall a bug that it would only disconnect.

roles/openstack/nova/tasks/cleanup.yml
```diff
@@ -0,0 +1,24 @@
+---
+# This file is used to cleanup OpenStack nova resources
+
+- name: Release floating IPs
+  os_floating_ip:
+      server: "{{ item.value.server | default(instance_name) }}"
+      state: absent
+  with_dict: "{{ provisioner.floating_ips }}"
```
yfried-redhat: why not `provisioner.instances`?

roles/openstack/nova/tasks/setup.yml
```diff
@@ -0,0 +1,47 @@
...
+
+- name: Create instances
+  os_server:
+      cloud: "{{ provisioner.cloud }}"
+      auto_ip: "{{ item.value.auto_ip | default(omit) }}"
+      floating_ip_pools: "{{ provisioner.floating_ip_pool | default(omit) }}"
+      config_drive: "{{ item.value.config_drive | default(config_drive) }}"
+      flavor: "{{ item.value.flavor | default(flavor_id) }}"
+      image: "{{ provisioner.image }}"
+      name: "{{ item.value.name | default(instance_name) }}"
```
yfried-redhat: `name: "{{ provisioner.prefix | default(random.uuid) }}-{{ item.value.name | default(instance_name) }}"

settings/provisioner/openstack/neutron/default.yml
```diff
@@ -0,0 +1,38 @@
...
+        name: '{{ random.uuid }}-router'
+        attach_subnets:
+            - "{{ !lookup provisioner.neutron.networks.external.name }}"
+            - "{{ !lookup provisioner.neutron.networks.management.name }}"
+        network: '10.8.184.0/22'
```
yfried-redhat: :-1: 
Major -1

This shouldn't be hard coded.
We've discussed a solution for finding available public networks. But it must have some default. possible:

- public
- nova
- omit

roles/openstack/nova/tasks/setup.yml
```diff
@@ -0,0 +1,47 @@
...
+      wait: "{{ item.value.wait | default(wait) }}"
+  register: nodes
+  with_dict: "{{ provisioner.topology }}"
+  when: provisioner.topology is defined
+
+- name: Create floating IPs
+  os_floating_ip:
+      cloud: "{{ provisioner.cloud }}"
+      fixed_address: "{{ item.value.fixed_address | default(omit) }}"
+      network: "{{ item.value.network | default(omit) }}"
```
yfried-redhat: this needs to match the router network

settings/provisioner/openstack/openstack.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+provisioner:
+    type: openstack
+
+distro:
```
yfried-redhat: this isn't used anywhere. image takes care of that. please remove this

bregman-arie: What exactly do you refer? provisioner.type + distro?

roles/openstack/nova/tasks/setup.yml
bregman-arie: It is. Please write specific line number.

settings/provisioner/openstack/neutron/default.yml
bregman-arie: What line?

roles/openstack/nova/tasks/setup.yml
bregman-arie: ack

roles/openstack/nova/tasks/cleanup.yml
bregman-arie: ack

settings/provisioner/openstack/neutron/default.yml
bregman-arie: This is following the same conventions as networks and subnets

settings/provisioner/openstack/openstack.yml
bregman-arie: Done.

settings/provisioner/openstack/neutron/default.yml
bregman-arie: Done.

plugins/filters/add_prefix.py
```diff
@@ -0,0 +1,12 @@
+from ansible import utils
+
+
+def add_prefix(value, prefix=''):
+    """Add string prefix to item."""
+    return prefix + value
+
+
+class FilterModule(object):
```
yfried-redhat: Could you please add a usage example to docstring?

requirements.txt
```diff
@@ -1,8 +1,7 @@
 ansible>=2.0.0
-python-novaclient<3.0.0
-python-neutronclient>=4.0.0
-PyYAML>=3.11
+clg>=2.0.0
 configure>=0.5
 colorlog>=2.6.1
-clg>=2.0.0
+PyYAML>=3.11
+shade
```
yfried-redhat: shade version?

roles/openstack/neutron/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
+# Network defaults
+
+external_network: False
+timeout_network: 180
+shared_network: False
```
yfried-redhat: Please remove (`external`|`shared`)`_network` default. there's no reasonable use case were user might wish to create all networks as external/shared.
If we have such an uncommon user, he will simply need to set his template explicitly.

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,26 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  os_router:
+      interfaces: "{{ item.value.attach_subnets | default(omit) }}"
+      name: "{{ item.value.name | default(router_name) }}"
```
yfried-redhat: shouldn't router name have prefix/random as well?

roles/openstack/neutron/defaults/main.yml
```diff
@@ -0,0 +1,17 @@
...
+
+enable_dhcp: True
+dns_servers:
+    - 208.67.222.222
+    - 8.8.8.8
+
+# Router defaults
+
+router_name: router1
+router_network: public
```
yfried-redhat: Please add a comment: "default name for external network"

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,26 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  os_router:
+      interfaces: "{{ item.value.attach_subnets | default(omit) }}"
```
yfried-redhat: I'm confused - don't you need the map filter to get the proper subnet names?

bregman-arie: Done

roles/openstack/neutron/defaults/main.yml
bregman-arie: No need. This is the defaults file. it should be clear.

roles/openstack/neutron/tasks/cleanup.yml
bregman-arie: Done.

roles/openstack/neutron/defaults/main.yml
yfried-redhat: It's not clear. router is connected to multiple network.
You could also change it to `router_external_network`

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  vars:
+      prefix: "{{ provisioner.prefix | default(uuid) }}-"
+  os_router:
+      interfaces: "{{ item.value.attach_subnets | map('add_prefix', prefix) | join(',') }}"
+      name: "{{ provisioner.prefix | default(uuid) }}-{{ item.value.name | default(router_name) }}"
```
yfried-redhat: `name: "{{ prefix }}-{{ item.value.name | default(router_name) }}"`

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  vars:
+      prefix: "{{ provisioner.prefix | default(uuid) }}-"
```
yfried-redhat: Why not move the prefix to defaults? and apply it to all names at once? instead of setting it as task var?
You could also put everything in a block and define block vars (same result).

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
...
+      name: "{{ prefix }}-{{ item.value.name | default(router_name) }}"
+      network: "{{ item.value.network | default(router_network) }}"
+      state: absent
+  with_dict: "{{ provisioner.neutron.routers }}"
+  when: provisioner.neutron.routers is defined
+  ignore_errors: yes
+
+- name: Cleanup subnets
+  os_subnet:
+      name: "{{ provisioner.prefix | default(uuid) }}-{{ item.value.name }}"
```
yfried-redhat: put prefix in defaults and avoid copy/pasting this all around the playbooks

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  vars:
+      prefix: "{{ provisioner.prefix | default(uuid) }}-"
+  os_router:
+      interfaces: "{{ item.value.attach_subnets | map('add_prefix', prefix) | join(',') }}"
+      name: "{{ prefix }}-{{ item.value.name | default(router_name) }}"
```
yfried-redhat: prefix already have "-" at the end.
How about: `{{ [prefix, item.value.name] | join("-") }}` ?

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
+---
+# This file is used to cleanup OpenStack neutron resources
+
+- name: Cleanup routers
+  vars:
+      prefix: "{{ provisioner.prefix | default(uuid) }}-"
+  os_router:
+      interfaces: "{{ item.value.attach_subnets | map('add_prefix', prefix) | join(',') }}"
```
yfried-redhat: Do you need `interfaces` and `network` for router cleanup? isn't router name enough?

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
...
+  when: provisioner.neutron.routers is defined
+  ignore_errors: yes
+
+- name: Cleanup subnets
+  os_subnet:
+      name: "{{ provisioner.prefix | default(uuid) }}-{{ item.value.name }}"
+      state: absent
+  with_dict: "{{ provisioner.neutron.subnets }}"
+  when: provisioner.neutron.subnets is defined
+  ignore_errors: yes
```
yfried-redhat: If we ignore errors, we need to reflect that cleanup failed some how.

my suggestion:
register each task (subnet, router, etc...) and run some sort of:

```
fail:
    msg: cleanup failed. Please review output to make sure no resources are leaking
when: subnets|failed or router|failed or ...
```

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -0,0 +1,30 @@
...
+
+- name: Cleanup networks
+  os_network:
+      name: "{{ provisioner.prefix | default(uuid) }}-{{ item.value.name }}"
```
yfried-redhat: same comment about reusing prefix

roles/openstack/neutron/tasks/main.yml
```diff
@@ -0,0 +1 @@
+- include: "{{ action }}.yml"
```
yfried-redhat: according to @tkammer we want to use "yaml" suffix instead of "yml"

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,60 @@
+---
+subparsers:
+    openstack:
+        help: Provision systems using Ansible OpenStack modules
+        groups:
+            - title: cloud
+              options:
+                  cloud:
+                      type: Value
+                      help: The cloud which the OpenStack modules will operate against
```
yfried-redhat: cloud needs more explanation. Most users don't know how to work with it.

roles/openstack/nova/tasks/setup.yml
```diff
@@ -0,0 +1,43 @@
...
+#      disk: "{{ item.value.disk | default(omit) }}"
+#      vcpus: "{{ item.value.vcpus | default(omit) }}"
+#      state: present
+#  register: flavors
+#  with_dict: "{{ provisioner.flavors }}"
+#  when: provisioner.flavors is defined  
+
+- name: Create instances
+  os_server:
+      cloud: "{{ provisioner.cloud }}"
```
yfried-redhat: Please use `default(omit)` for cloud value to allow users to source "keystonerc" files. This is a great functionality in shade in the UX front.

settings/provisioner/openstack/openstack.spec
yfried-redhat: We discussed adding the "glance image-list" command to the `--image` help text so users can know which flavors are available for them.

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,60 @@
...
+subparsers:
+    openstack:
+        help: Provision systems using Ansible OpenStack modules
+        groups:
+            - title: cloud
+              options:
+                  cloud:
+                      type: Value
+                      help: The cloud which the OpenStack modules will operate against. Cloud setup instructions: http://docs.openstack.org/developer/os-client-config/#config-files
+                      required: yes
```
yfried-redhat: Please remove "required" and add to help that "you can source valid keystonerc files instead or define OS_* envvars instead"

settings/provisioner/openstack/openstack.spec
```diff
@@ -0,0 +1,60 @@
...
+            - title: prefix
+              options:
+                  prefix:
+                      type: Value
+                      help: An prefix which would be contacted to each provisioned resource
```
yfried-redhat: help: "If not provided, all resource names will be prefixed with a common random string"

roles/openstack/nova/tasks/setup.yml
```diff
@@ -0,0 +1,44 @@
...
+      flavor: "{{ item.value.flavor | default(flavor_id) }}"
+      image: "{{ provisioner.image }}"
+      name: "{{ provisioner.prefix | default(random.uuid) }}-{{ item.value.name }}"
+      nics: "{{ networks.results | map(attribute='id') | list() |map('add_prefix', 'net-id=') | list() }}"
+      key: "{{ item.value.key | default(omit) }}"
+      key_name: "{{ item.value.key_name | default(omit) }}"
+      state: present
+      wait: "{{ item.value.wait | default(wait) }}"
+  register: nodes
+  with_dict: "{{ provisioner.topology }}"
```
yfried-redhat: since you changed the spec, the dict is now at `provisioner.topology.nodes` (the same applies for neutron)

rhosqeauto:

rhosqeauto:

roles/openstack/nova/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
+# General defaults
+
+prefix: "{{ provisioner.prefix | default(random.uuid) }}"
+
+# Instances defaults
+
+auto_ip: no
+config_drive: no
+instance_image: centos-minimal
```
mpryc: Why don't we use rhel-guest-image-7.2 ?

roles/openstack/nova/tasks/cleanup.yml
```diff
@@ -0,0 +1,45 @@
...
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      server: "{{ prefix }}*"
+  register: servers_details
+- debug: var=servers_details
+
+- name: Set floating IP address
+  set_fact:
+      floating_ip_address: "{{ item.value[0].interface_ip }}"
+  with_dict: servers_details.ansible_facts
+  ignore_errors: yes
```
mpryc: Why there is need to set the floating ip address when cleaning up?

roles/openstack/neutron/tasks/main.yml
mpryc: Adding some comment on top of this file explaining what this task is about is more then welcome.

plugins/filters/add_prefix.py
```diff
@@ -0,0 +1,17 @@
...
+    -> ['happy hanukkah', 'happy easter', 'happy pessah']
+    """
+
+    return prefix + value
+
+
+class FilterModule(object):
+
+    def filters(self):
+        return {'add_prefix': add_prefix}
```
mpryc: Please keep the convention same as other filters:

return {
              'add_prefix': add_prefix
}

plugins/filters/add_prefix.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def add_prefix(value, prefix=''):
+    """Add string prefix to item.
+
+    {{ ['hanukkah', 'easter', 'pessah'] | map('add_prefix', 'happy ' }}
+    -> ['happy hanukkah', 'happy easter', 'happy pessah']
+    """
+
+    return prefix + value
```
mpryc: Use of:

return '%s%s' % (prefix, value)

roles/openstack/nova/defaults/main.yml
bregman-arie: This is the defaults file. it include default values for various variables.
For image I don't think we can assume all users have 'rhel-guest-image-7.2' on their clouds.
I think it should be required from the user to provide value for it. default cannot be assumed here.

mpryc: Can we assume that all users have centos-minimal? Don't think so... so we should use default that suits us the best which isn't centos-minimal imo.

plugins/filters/add_prefix.py
bregman-arie: Can you explain why this way is better?

mpryc: If an object defines __str__ but it's not type of String you will get error using +, concentrating strings using above will properly do so. Just one example. Performance wise is not relevant here as it's not too many concentrated strings, but just in general it's + is O(n^2) and above is O(n).

roles/openstack/nova/defaults/main.yml
bregman-arie: I think we can assume more users will have centos-minimal or the ability to upload it.

roles/openstack/nova/tasks/cleanup.yml
bregman-arie: "I am unsure if we should ignore_errors here?" I don't think we can assume every node has '"{{ item.value[0].interface_ip }}"'

"What happens if the floating IP is not set, will the Release floating IP work?" it will simply be skipped.

"Also this is used only in Release floating IPs, so why there is separate action to get just the interface_ip?" not sure I'm following. how do you suggest to retrieve this value without separate task?

"Isn't it easier to move this to the Release floating IPs part and remove whole section?"
I'm open to suggestions. Please post the code you think should replace this implementation.

plugins/filters/add_prefix.py
bregman-arie: OK, I read Guido's essay and looks like join is the best way. will change it. thanks.

plugins/filters/add_prefix.py
bregman-arie: Done.

rhosqeauto:

rhosqeauto:

roles/openstack/nova/tasks/cleanup.yml
mpryc: OK so how about:

name: Register nova servers details
to:
name: Cleanup: Set nova servers details

name: Set floating IP address
to:
name: Cleanup - Set floating IP address

name: Release floating IPs
to:
name: Cleanup floating IPs

From ansible documentation "registered variables are just like facts" so we shouldn't distinguish them in the naming of those playbooks.

bregman-arie: I actually thought you meant something else entirely..
You think we simply need to change the name of the tasks?

mpryc: I meant something else entirely, but after your explanation I understand why it's implemented this way and I am fine with it, so just proposing name change to be more meaningful.

rhosqeauto:

rhosqeauto:

yfried-redhat:
In next commit:
Inventory:
- [ ] write inventory to hosts file
- [ ] hosts file should have prefix in name: `"hosts-{{ prefix }}"`
key file:
- [ ] user must provide `--key-file` (can use ~/.ssh/id_rsa as default)
- [ ] user can provide `--key-name`
- [ ] if `--key-name` is missing:
    - generate public key from from file
    - optional - look for pubkey in existing keys on cloud
    - upload pubkey to cloud (don't need to delete it on cleanup)    

- [ ] floating ip are recreated on 2nd run - check this out

yfried-redhat:
Why does floating ip show as "changed" when it's not:

```
(InfraRed)[yfried ~/workspace/git/InfraRedCI] # neutron floatingip-list                                                                                                                                                    
+--------------------------------------+------------------+---------------------+--------------------------------------+
| id                                   | fixed_ip_address | floating_ip_address | port_id                              |
+--------------------------------------+------------------+---------------------+--------------------------------------+
| 2eca16a6-2967-48cb-b570-2a772e6e2b4b | 192.168.1.13     | 10.8.180.160        | 1cd46ef7-bc33-417d-bc17-9e24c48220d6 |
+--------------------------------------+------------------+---------------------+--------------------------------------+

TASK [openstack/nova : Create floating IPs] ************************************
                      [[ previous task time: 0:00:06.410308 = 6.41s / 20.72s ]]
ESTABLISH LOCAL CONNECTION FOR USER: yfried
localhost EXEC /bin/sh -c '( umask 22 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1460010726.03-151012890134258 `" && echo "` echo $HOME/.ansible/tmp/ansible-tmp-1460010726.03-151012890134258 `" )'
localhost PUT /tmp/tmp6LNCf1 TO /home/yfried/.ansible/tmp/ansible-tmp-1460010726.03-151012890134258/os_floating_ip
localhost EXEC /bin/sh -c 'LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 python /home/yfried/.ansible/tmp/ansible-tmp-1460010726.03-151012890134258/os_floating_ip; rm -rf "/home/yfried/.ansible/tmp/ansible-tmp-1460010726.03-151012890134258/" > /dev/null 2>&1'
changed: [localhost] => (item={'value': {'name': 'controller', 'disks': {'disk1': {'path': '/var/lib/libvirt/images', 'dev': '/dev/vda', 'size': '30G'}}, 'amount': 1, 'groups': ['controller', 'openstack_nodes'], 'memory': '8192', 'os': {'type': 'linux', 'variant': 'rhel7'}, 'cpu': '4', 'network': {'interfaces': {'management': {'label': 'eth0'}, 'data': {'label': 'eth1'}, 'external': {'label': 'eth2'}}, 'floating_ip_network': 'management'}}, 'key': 'controller'})


(InfraRed)[yfried ~/workspace/git/InfraRedCI] # neutron floatingip-list                                                                                                                                                    
+--------------------------------------+------------------+---------------------+--------------------------------------+
| id                                   | fixed_ip_address | floating_ip_address | port_id                              |
+--------------------------------------+------------------+---------------------+--------------------------------------+
| 2eca16a6-2967-48cb-b570-2a772e6e2b4b | 192.168.1.13     | 10.8.180.160        | 1cd46ef7-bc33-417d-bc17-9e24c48220d6 |
+--------------------------------------+------------------+---------------------+--------------------------------------+

```

yfried-redhat:
Cleanup should fail at the end if tasks failed

roles/openstack/nova/tasks/cleanup.yml
```diff
@@ -1,37 +1,15 @@
...
+  with_dict: provisioner.nodes
 
-- name: Cleanup instances
-  os_server:
+- name: Cleanup key pairs
+  os_keypair:
       cloud: "{{ provisioner.cloud | default(omit) }}"
-      image: "{{ item.value.image | default(instance_image) }}"
-      name: "{{ prefix }}{{ item.value.name }}"
+      name: "{{ provisioner.key_file | basename }}"
```
yfried-redhat: This will delete keys if even if they existed on the server, wouldn't it?

yfried-redhat: Resources created should have prefix, shouldn't they?

roles/openstack/nova/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,36 @@
...
+- name: Create instances
+  os_server:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      auto_ip: "{{ node_type.value.auto_ip | default(auto_ip) }}"
+      config_drive: "{{ node_type.value.config_drive | default(config_drive) }}"
+      flavor: "{{ node_type.value.flavor | default(flavor_id) }}"
+      image: "{{ provisioner.image }}"
+      name: "{{ prefix }}{{ node_type.value.name }}-{{ item }}"
+      nics: "{{ networks.results | map(attribute='id') | list() |map('add_prefix', 'net-id=') | list() }}"
+      key: "{{ node_type.value.key | default(omit) }}"
```
yfried-redhat: keys aren't unique to node. you should use provisioner.key_file

roles/openstack/nova/tasks/setup.yml
```diff
@@ -14,30 +14,19 @@
...
 
-- name: Create floating IPs
-  os_floating_ip:
+- name: Add keypair
+  os_keypair:
       cloud: "{{ provisioner.cloud | default(omit) }}"
-      network: "{{ external_network_id }}"
-      internal_network: "{{ prefix}}{{ item.value.network.floating_ip_network }}"
-      server: "{{ prefix }}{{ item.value.name }}"
+      name: "{{ provisioner.key_file | basename }}"
```
yfried-redhat: add prefix to key

settings/provisioner/openstack/openstack.spec
```diff
@@ -14,6 +14,18 @@ subparsers:
...
                       help: An prefix which would be contacted to each provisioned resource. An random prefix would be generated in case this value is not specified.
+            - title: key_file
+              options:
+                  key_file:
+                      type: Value
+                      help: The key file that would be uploaded to nova and injected into VMs upon creation.
+                      default: ~/.ssh/id_rsa
+            - title: key_name
+              options:
+                  key_name:
```
yfried-redhat: If we are always deleting keys from openstack, it doesn't make sense for user to add his key here.
Let's remove `key_name` and always generate the keys from file.

roles/openstack/nova/tasks/setup_nodes.yml
bregman-arie: Done.

settings/provisioner/openstack/openstack.spec
bregman-arie: I decided to remove keypair cleanup

roles/openstack/neutron/tasks/setup.yml
```diff
@@ -52,3 +52,25 @@
...
+
+- name: Create security groups
+  os_security_group:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      name: "{{ prefix }}{{ item.value.name | default(security_group_name) }}"
+      state: present
+  register: security_groups
+  with_dict: "{{ provisioner.neutron.security_groups }}"
+  when: provisioner.neutron.security_groups is defined
+
```
yfried-redhat: Where do we have the secgroup dict? Is it missing from openstack.yaml file?
Note - I don't think we need to expose this in spec ATM

roles/openstack/neutron/tasks/cleanup.yml
```diff
@@ -26,4 +28,18 @@
...
+  register: networks
   ignore_errors: yes
+
+- name: Cleanup security groups
+  os_security_group:
+      cloud: "{{ provisioner.cloud | default(omit) }}" 
+      name: "{{ prefix }}{{ item.value.name | default(security_group_name) }}" 
+      state: absent
+  register: security_groups
+  with_dict: "{{ provisioner.neutron.security_groups }}"
```
yfried-redhat: This needs to be commented out until 2.1, and the same `ignore_errors` flag

roles/openstack/nova/tasks/cleanup_nodes.yml
```diff
@@ -0,0 +1,34 @@
...
+  os_server:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      name: "{{ prefix }}{{ node_type.value.name }}-{{ item }}"
+      state: absent
+  with_sequence: "count={{ node_type.value.amount }}"
+  when: provisioner.nodes is defined
+  register: instances
+  ignore_errors: yes
+
+- name: Cleanup floating IPs
```
yfried-redhat: This task is redundant until we change the module to delete FLIPs instead of just releasing them.

roles/openstack/nova/tasks/setup.yml
```diff
@@ -14,30 +14,19 @@
...
-      key_name: "{{ item.value.key_name | default(omit) }}"
-      state: present
-      wait: "{{ item.value.wait | default(wait) }}"
-  register: nodes
-  with_dict: "{{ provisioner.nodes }}"
-  when: provisioner.nodes is defined
+# This is needed for os_keypair as it uses public key
+- name: Retrieve public key from private key
+  shell: "ssh-keygen -y -f {{ provisioner.key.file }}"
+  register: public_key_content
```
yfried-redhat: `when: not provisioner.key.name`

roles/openstack/nova/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: Create floating IPs
+  os_floating_ip:
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+      network: "{{ external_network_id }}"
+      internal_network: "{{ prefix}}{{ node_type.value.network.floating_ip_network }}"
+      server: "{{ prefix }}{{ node_type.value.name }}-{{ item }}"
+      state: present
+  register: floating_ips
+  with_sequence: "count={{ node_type.value.amount }}"
+  when: provisioner.nodes is defined
```
yfried-redhat: The reason you don't have the inventory file is because you never add the created hosts to inventory:
```yaml
- os_server_facts:
    cloud: <>
    server: "{{ prefix }}{{ node_type }}*"
  register: servers

```

Will grab all servers and you can run add_host on it's output.
remember to save the item.public_ip as the ansible address.

roles/openstack/neutron/tasks/cleanup.yml
bregman-arie: Done.

roles/openstack/nova/tasks/cleanup_nodes.yml
bregman-arie: Done.

roles/openstack/nova/tasks/setup_nodes.yml
bregman-arie: Done

roles/openstack/nova/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,48 @@
...
+
+- name: Gather created servers facts
+  os_server_facts:
+      server: "{{ prefix }}{{ node_type.value.name }}*"
+      cloud: "{{ provisioner.cloud | default(omit) }}"
+  register: servers
+- debug: var=servers
```
yfried-redhat: remove this please

roles/openstack/nova/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,48 @@
...
+      
+- name: Add created servers to inventory
+  add_host:
+      name:"{{ prefix }}{{ node_type.value.name}}-{{item}}"
+      ansible_ssh_host="{{ servers.ansible_facts.openstack_servers[0].interface_ip }}"
```
yfried-redhat:
```yaml
     name: "{{ prefix }}{{ node_type.value.name}}-{{item}}"
     ansible_host: "{{ servers.ansible_facts.openstack_servers[0].public_ip }}"
     groups: "{{ provisioner.nodes.node_type.groups }}"
     ansible_ssh_private_key_file: "{{ provisioner.key.file }}"
     ansible_ssh_user: cloud-user
```

roles/openstack/nova/tasks/setup_nodes.yml
bregman-arie: Done

roles/openstack/nova/tasks/setup_nodes.yml
bregman-arie: why to use fixed value as 'cloud-user' ?

yfried-redhat: You are right. But it's a good default

Regards
Yair

PR #126, opened 2016-03-22
RHOSINFRA-80 Execute Ansible 2

yfried-redhat:
Converts execute to use 2.0

### DIFF
--- a/cli/execute.py
+++ b/cli/execute.py
@@ -41,7 +41,6 @@ def ansible_playbook(playbook, args, inventory="local_hosts"):
 
     LOG.info("Executing Playbook: %s" % playbook)
 
-    variable_manager = VariableManager()
     loader = DataLoader()
 
     # ------------------ Mocking ansible-playbook cli input ------------------
@@ -82,14 +81,12 @@ def ansible_playbook(playbook, args, inventory="local_hosts"):
     options = namedtuple('Options', hacked_options.keys())(**hacked_options)
 
     passwords = dict(vault_pass='secret')
+    variable_manager = VariableManager()
+    variable_manager.extra_vars = args["settings"]
     inventory = Inventory(loader=loader, variable_manager=variable_manager,
                           host_list=inventory)
     variable_manager.set_inventory(inventory)
 
-    loader = DataLoader()
-
-    variable_manager = VariableManager()
-    variable_manager.extra_vars = args["settings"]
     pbex = PlaybookExecutor(playbooks=[path_to_playbook], inventory=inventory,
                             variable_manager=variable_manager, loader=loader,
                             options=options, passwords=passwords)
--- a/cli/execute.py
+++ b/cli/execute.py
@@ -1,10 +1,14 @@
 from os import path
 
-import ansible.color
-import ansible.inventory
-import ansible.playbook
-import ansible.utils
-from ansible import callbacks
+from collections import namedtuple
+import ansible.constants
+from ansible.parsing.dataloader import DataLoader
+from ansible.vars import VariableManager
+from ansible.inventory import Inventory
+from ansible.executor.playbook_executor import PlaybookExecutor
+# from ansible.playbook.play_context import PlayContext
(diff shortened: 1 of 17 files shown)


yfried-redhat:
@tkammer @bregman-arie @obaranov @aopincar please review

yfried-redhat:
retest

yfried-redhat:
retest

yfried-redhat:
retest

cli/execute.py
```diff
@@ -19,32 +21,6 @@
     CONF.get('defaults', 'playbooks')), "Bad path to playbooks"
```
obaranov: Why do we need that assert here? It looks like we are forcing user to use "playbooks" folder. In that case we don't need configuration option for playbooks.

cli/execute.py
```diff
@@ -52,6 +28,10 @@ def ansible_playbook(playbook, args, inventory="local_hosts"):
      :args: all arguments passed for the playbook
      :inventory: the inventory file to use, default: local_hosts
     """
+    verbosity = args["verbose"]
+    display = Display(verbosity=verbosity)
+    import __main__ as main
+    setattr(main, "display", display)
```
obaranov: Why do we need that? If 'display' is required for some one else, then maybe it should be passed to the method as input argument or configured in the 'config' module

cli/execute.py
```diff
@@ -59,96 +39,57 @@ def ansible_playbook(playbook, args, inventory="local_hosts"):
...
-        raise exceptions.IRPlaybookFailedException(
-            playbook, "Failed hosts: %s" % failed_hosts)
-    if len(unreachable_hosts) > 0:
-        raise exceptions.IRPlaybookFailedException(
-            playbook, "Unreachable hosts: %s" % unreachable_hosts)
+    options = namedtuple('Options', hacked_options.keys())(**hacked_options)
+
+    passwords = dict(vault_pass='secret')
+    variable_manager = VariableManager()
+    variable_manager.extra_vars = args["settings"]
```
obaranov: Is the 'settings' key is standard to ansible 2.0. I think we need to document the structure of the args parameter.

cli/execute.py
```diff
@@ -59,96 +39,57 @@ def ansible_playbook(playbook, args, inventory="local_hosts"):
...
+    inventory = Inventory(loader=loader, variable_manager=variable_manager,
+                          host_list=inventory)
+    variable_manager.set_inventory(inventory)
+
+    pbex = PlaybookExecutor(playbooks=[path_to_playbook], inventory=inventory,
+                            variable_manager=variable_manager, loader=loader,
+                            options=options, passwords=passwords)
+    results = pbex.run()
+    if results:
+        raise exceptions.IRPlaybookFailedException(playbook)
```
obaranov: Just a thought. Maybe it worth to return results at the end of the method?

yfried-redhat: Results is a return code (0/2/3).
This just means moving the "if results - raise..." to a different module

yfried-redhat:
@bregman-arie If you wan't to use ansible 2 in your work, pull this to your branch.

bregman-arie:
When will it be merged?

cli/execute.py
yfried-redhat: That's how ansible 2 sets verbosity.
It's a hack, because ansible-playbook isn't fully supported to be consumed from python API instead of CLI.
I am looking for a better way. So far, that's all I've got

cli/execute.py
yfried-redhat: Yep. it's a leftover from old code. will remove it.

yfried-redhat:
@bregman-arie It was supposed to merge, but I'm seeing some problems with installer. probably some incompatible playbooks. Since you aren't using any of those playbooks, just pull from this branch and it will automatically install and execute latest ansible 2 release instead of 1.9

cli/execute.py
yfried-redhat: agreed

PR #127, opened 2016-03-23
RHOSINFRA-90 Init only chosen subparser's args

yfried-redhat:


obaranov:
Looks good for me.  no comments. Merging

PR #139, opened 2016-04-01
[RHOSINFRA-70] Image settings

obaranov:
Replace the image* arguments for provisioner with distro one. 

@tkammer, @yfried-redhat please review. 

This change requires also CI private change: https://gitlab.cee.redhat.com/rhosqeauto/InfraRedCI/merge_requests/15


obaranov:
POC job: http://10.8.180.145:8080/job/poc-ir-obaranov-distro/4/console

settings/provisioner/distro/rhel-7.2.yml
```diff
@@ -0,0 +1,8 @@
+image:
+  file: rhel-guest-image-7.2-20160302.0.x86_64.qcow2
+  server: "{{ !lookup private.images.rhel.7.2.remote_file_server }}"
+name: rhel
```
yfried-redhat: Why do we need this data?
We are using images to specify the distro, so we don't need to hold it in yaml

```yaml
name: rhel
full_version: "7.2"
 version:
   major: '7'
   minor: '2'
```

obaranov: It's just a common information about the distro which is currently not used. But the distro yaml can be used by different apps. 

I can remove it but I think we would need that data later.

cli/spec.py
```diff
@@ -130,6 +130,25 @@ def resolve_value(self, arg_name, defaults=None):
         self.value = utils.load_yaml(self.value, search_first)
 
 
+class SpecYamlArgument(YamlFileArgument):
```
yfried-redhat: This isn't needed at this moment. see comment about distro file

settings/provisioner/distro/rhel-7.2.yml
```diff
@@ -0,0 +1,8 @@
+image:
```
yfried-redhat: This file is specific for virsh, so please keep it in virsh settings

obaranov: Let's discuss that with @tkammer. In the ticket I had the requirement to have distro outside virsh provisioner. The idea was to have something similar to khaleese were we have distro yaml file with all the information about specific distro. And that infroamtion were used by different modules.

yfried-redhat:
-1
Please don't merge this. let's discuss this in meeting

settings/provisioner/virsh/virsh.spec
```diff
@@ -17,10 +17,10 @@ subparsers:
...
                       help: "User's SSH key"
                       default: ~/.ssh/id_rsa
-            - title: distro
+            - title: image
               options:
-                  distro:
-                      type: CommonSpecYamlFile
+                  image:
+                      type: YamlFile
                       help: The RHEL distribution version
```
yfried-redhat: Help message is wrong here.

settings/provisioner/virsh/virsh.spec
```diff
@@ -19,13 +19,9 @@ subparsers:
...
-                  image-file:
-                      type: Value
-                      help: An image to provision the host with
-                      required: yes
-                  image-server:
-                      type: Value
-                      help: Base URL of the image file server
+                  image:
+                      type: YamlFile
+                      help: The RHEL distribution version
```
yfried-redhat: The help text is wrong here. please explain that this is the path to a file containing the image and the server details

obaranov:
Fixed both public an private pull request.

PR #142, opened 2016-04-03
Add shade to requirements

bregman-arie:
Shade is used by OpenStack modules.


rhosqeauto:

yfried-redhat:
This should be part of the branch that introduces the new openstack modules. Until we actually start using them, there's no need to add them to requirements.
Unlike gerrit you can have multiple commits in the same branch

PR #143, opened 2016-04-04
[RHOSINFRA-96] Fix 'choices' issue in spec parser

obaranov:
Now we can use choices for Value arguments. 

@yfried-redhat @aopincar, please review.

### DIFF
--- a/cli/spec.py
+++ b/cli/spec.py
@@ -1,4 +1,5 @@
 import ConfigParser
+from functools import total_ordering
 
 import clg
 import os
@@ -19,6 +20,7 @@ BUILTINS_REPLACEMENT = {
 TRIM_PARAMS = ['default', 'required']
 
 
+@total_ordering
 class ValueArgument(object):
     """
     Default argument type for InfraRed Spec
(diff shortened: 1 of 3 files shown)


cli/spec.py
```diff
@@ -1,4 +1,5 @@
 import ConfigParser
+from functools import total_ordering
```
yfried-redhat: Those this means we need to add functools to requirements? If so, please do

cli/spec.py
```diff
@@ -108,6 +110,27 @@ def resolve_value(self, arg_name, defaults=None):
...
+    def __eq__(self, other):
+        """
+        Checks if other value is equal to the current value.
+        """
+        if isinstance(other, ValueArgument):
+            return self.value == other.value
+        else:
+            return self.value == other
+
+    def __lt__(self, other):
```
yfried-redhat: How does this apply to YamlFileArgument?

obaranov: self.value for YamlFileArgument is the dict. So this will compare two dict in case for YamlFileArgument
Also I think we cannot use choices for YamlFileArgument's because we transfrom the actual argument value by loading yaml file.

cli/spec.py
obaranov: Python2.7 comes with the functools module included.

yfried-redhat:
2 small concerns. Please fix. tnx

PR #147, opened 2016-04-06
Add the option to run smoke tests only on tempest

arxcruz:
This add the option to run smoke tests with --tempest-tests=smoke


rhosqeauto:

rhosqeauto:

yfried-redhat:
@arxcruz I'm closing this becuase:

1. On the ticket I've explained why we don't want nightly jobs to run whole smoke suite.
1. If required, this can be easily achieved with `--extar-vars` and without a change to the repo
1. Current ksgen and settings/tester will be removed in the near future.

PR #156, opened 2016-04-11
[RHOSINFRA-130]  Refactor Lookup mechanism & removes the use in the 'configure' module

aopincar:
Implements new Lookup mechanism
Modifies unittests according the changes
Removes un-needed YAML constructors
Removes the use in the 'configure' module

Resolves #RHOSINFRA-130


yfried-redhat:
retest openstack

yfried-redhat:
retest virsh

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
-        ret = os.getenv(var, seq[1])  # second item is default val
+    try:
+        with open(settings_file) as f_obj:
+            loaded_yml = yaml.load(f_obj)
 
-        # third item is max. length
-        if len(seq) == 3:
-            ret = _limit_chars(ret, seq[2])
-        return ret
+        # Handling case of empty file
```
tkammer: Why are we handling a case of empty files? is this a valid choice?

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
 
-def _limit_chars(_string, length):
-    length = int(length)
-    if length < 0:
-        raise exceptions.IRException('length to crop should be int, not ' +
-                                     str(length))
+    @classmethod
+    def from_yaml(cls, loader, node):
+        # Create & save references to Placeholder objects
+        placeholder = Placeholder(str(node.start_mark))
```
tkammer: This line is redundant..

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
+        self.message = message
+        self.file_path = None
 
-def _limit_chars(_string, length):
-    length = int(length)
-    if length < 0:
-        raise exceptions.IRException('length to crop should be int, not ' +
-                                     str(length))
+    @classmethod
+    def from_yaml(cls, loader, node):
```
tkammer: why is this called "from_yaml" if this returns a "placeholder" object?

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
 
+        chars = string.ascii_lowercase + string.digits
+        return ''.join(random.choice(chars) for _ in range(int(node.value)))
 
-@configure.Configuration.add_constructor('join')
-def _join_constructor(loader, node):
-    seq = loader.construct_sequence(node)
-    return ''.join([str(i) for i in seq])
 
+class Placeholder(yaml.YAMLObject):
```
tkammer: Please rethink / explain this whole class..

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
+    @classmethod
+    def from_yaml(cls, loader, node):
+        # Create & save references to Placeholder objects
+        placeholder = Placeholder(str(node.start_mark))
+        cls.placeholders_list.append(placeholder)
+        return placeholder
 
-    return _string[:length]
+    @classmethod
+    def to_yaml(cls, dumper, node):
```
tkammer: again, no idea what is the connection between "to_yaml" and it raising an exception

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
+        message = re.sub("<string>", node.file_path, node.message)
+        raise exceptions.IRPlaceholderException(message)
 
 
-@configure.Configuration.add_constructor('limit_chars')
-def _limit_chars_constructor(loader, node):
+def load(settings_file, update_placeholders=True):
```
tkammer: doesn't the regular "load" function does this already? why create a dedicated function?

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
+        raise exceptions.IRYAMLConstructorError(e, settings_file)
 
-    settings = None
 
-    def __init__(self, key, old_style_lookup=False):
-        self.key = key
-        if old_style_lookup:
-            self.convert_old_style_lookup()
-        self.handling_nested_lookups = False
+def _dict_flattener(ord_dict):
```
tkammer: Why do we need this?

cli/yamls.py
```diff
@@ -1,358 +1,217 @@
...
-        if key not in dic:
-            if isinstance(key, str) and key.isdigit():
-                key = int(key)
-            elif isinstance(key, int):
-                key = str(key)
 
-        try:
-            if keys:
-                return self.dict_lookup(keys, dic[key])
+def _dict_inflator(flattened_dict):
```
tkammer: again, why?

yfried-redhat:
retest openstack

cli/yamls.py
aopincar: Valid yes, reasonable no!
A proper IRException is being raised now if loading an empty file.
Fixed.

cli/yamls.py
aopincar: Why?
Creating a Placeholder object, and save a reference to it in the placeholders list...

cli/yamls.py
aopincar: When loading a file using yaml.load, the value is replaced with Placeholder object..
When dumping the file using yaml.dump, if there are any Placeholder objects, an exception will be raised with the exact trace to the yaml file where this Placeholder came from.
This is important when having more than one placeholder and/or merging many yaml files

cli/yamls.py
aopincar: We're sub-classing YAMLObject in order to create YAML constructors

cli/yamls.py
aopincar: Add the final stage, when dumping the merging result of all settings file, if not all Placeholder object are replaced, an exception will be raised.

cli/yamls.py
aopincar: In order to save the path to the file of each file which has placeholder tag

cli/install.py
```diff
@@ -86,24 +86,17 @@ def main():
...
-    # todo(yfried): ospd specific
-    cli.yamls.Lookup.settings = cli.yamls.Lookup.settings.merge(settings_dict)
-
-    cli.yamls.Lookup.in_string_lookup()
-
-    LOG.debug("Dumping settings...")
-
-    output = yaml.safe_dump(cli.yamls.Lookup.settings,
-                            default_flow_style=False)
+    all_settings = utils.load_settings_files(settings_files)
```
yfried-redhat: This module is not used. should be deprecated

cli/yamls.py
```diff
@@ -1,358 +1,243 @@
...
+    yaml_tag = u'!random'
+    yaml_dumper = yaml.SafeDumper
 
-def random_generator(size=32, chars=string.ascii_lowercase + string.digits):
-    import random
-
-    return ''.join(random.choice(chars) for _ in range(size))
+    @classmethod
+    def from_yaml(cls, loader, node):
+        import random
```
yfried-redhat: why do we need to import here and not in the usual place?

cli/yamls.py
yfried-redhat: @tkammer the simple implementation of placeholder tag didn't return the object's key and yaml file in Exception.
You requested a better output for the user in early stages of the project.

However, I'm not sure if we use (or should use) the "placeholder" tag at all.

cli/yamls.py
obaranov: @aopincar if we do not allow empty yaml files, then we need to fix the installer/ospd/workarounds/rhos-8.0.yml file. This file is just a placeholder for future workarounds. 
Otherwise ospd installer for 8.0 will fail.

cli/yamls.py
```diff
@@ -111,7 +111,7 @@ def _dict_flattener(ord_dict):
     def dict_builder(_dic, parent_key=''):
         for key, value in _dic.iteritems():
             if parent_key:
-                key = parent_key + '.' + key
+                key = parent_key + '.' + str(key)
```
arxcruz: I would like to see a test with an integer key. So in the future if someone change this function, ensure that still expecting the same result.

cli/yamls.py
aopincar: @tkammer your opinion is needed here!!!

tkammer: remove rhos-8.0.yml

aopincar: @obaranov are you taking care of removing this?

obaranov: @aopincar I can do that. Create new PR for that?

aopincar: @obaranov yes please. but lets wait for this patch to be merged before doing that.

cli/yamls.py
```diff
@@ -330,29 +57,126 @@ def to_yaml(cls, dumper, node):
...
-    except configure.ConfigurationError as confError:
-        # fix for the empty yml files
-        if confError.message == 'unconfigured':
-            return {}
-        else:
-            raise
+        raise exceptions.IRYAMLConstructorError(e, settings_file)
+
+
+def _get_most_nested_lookups(string_value):
```
tkammer: Remove one line methods - in general, try not building method / classes that are nothing but a "return"

cli/yamls.py
```diff
@@ -330,29 +57,126 @@ def to_yaml(cls, dumper, node):
...
+
+
+def replace_lookup(lookups_dict):
+    """
+    Replaces all lookup pattern in a given dictionary (lookups_dict) with
+    the corresponding values
+
+    :param lookups_dict: Dictionary with data to replace
+    """
+    def yaml_walk(yaml_content, key_path=None):
```
tkammer: Please remove this inner method as there is no extra logic in having it as an inned method

cli/yamls.py
```diff
@@ -330,29 +57,126 @@ def to_yaml(cls, dumper, node):
...
+
+def dict_get(dic, key):
+    """
+    Gets the value of a nested key in a dictionary
+
+    :param dic: Dictionary containing the needed data
+    :param key: String representing full key path (Example: 'key.sub1.sub2')
+    :return: Value of a given key in the dictionary
+    """
+    def _dict_get(_dic, current_key, *keys):
```
tkammer: Please remove this inner method as there is no extra logic in having it as an inned method

cli/yamls.py
```diff
@@ -330,29 +57,126 @@ def to_yaml(cls, dumper, node):
...
+    Replaces all lookup pattern in a given dictionary (lookups_dict) with
+    the corresponding values
+
+    :param lookups_dict: Dictionary with data to replace
+    """
+    def yaml_walk(yaml_content, key_path=None):
+        an_iter = enumerate(yaml_content) if isinstance(yaml_content, list) \
+            else yaml_content.iteritems()
+
+        for idx_key, value in an_iter:
```
tkammer: please use full words rather than abbreviations (use: index_key, value rather than idx_key)

cli/yamls.py
```diff
@@ -330,29 +57,126 @@ def to_yaml(cls, dumper, node):
...
+
+def replace_lookup(lookups_dict):
+    """
+    Replaces all lookup pattern in a given dictionary (lookups_dict) with
+    the corresponding values
+
+    :param lookups_dict: Dictionary with data to replace
+    """
+    def yaml_walk(yaml_content, key_path=None):
+        an_iter = enumerate(yaml_content) if isinstance(yaml_content, list) \
```
tkammer: please don't instantiate a variable that can potentially be two types.
You should check if it's a list and go over it to look for a !lookup.
Once you got it, call the 'walk' with that as yaml_content

PR #158, opened 2016-04-12
Don't fail os_floating_ip on 2nd run

yfried-redhat:
Check that existing floater matches requirements

### DIFF
--- a/library/os_floating_ip.py
+++ b/library/os_floating_ip.py
@@ -182,6 +182,29 @@ def main():
                         fixed_address = address['ip_address']
                         break
 
+            # If f_ip already assigned to server, check that it matches
+            # requirements.
+            fip_address = cloud.get_server_public_ip(server)
+            if fip_address:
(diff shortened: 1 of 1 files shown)


library/os_floating_ip.py
```diff
@@ -182,6 +182,29 @@ def main():
                         fixed_address = address['ip_address']
                         break
 
+            # If f_ip already assigned to server, check that it matches
+            # requirements.
```
bregman-arie: "already assigned to the server, check that it matches the requirements."

library/os_floating_ip.py
```diff
@@ -182,6 +182,29 @@ def main():
...
+            fip_address = cloud.get_server_public_ip(server)
+            if fip_address:
+                f_ip = _get_floating_ip(cloud, fip_address)
+                if all([fixed_address, f_ip.fixed_ip_address == fixed_address,
```
bregman-arie: Why are you checking that fixed_address is equal to fixed_address?

PR #159, opened 2016-04-12
Cleanup floating ips

yfried-redhat:
Pull latest module so it can purge floating ips.
Add option to delete floating ip by server and not by address.


yfried-redhat:
retest virsh

bregman-arie:
I approve this message!

PR #173, opened 2016-04-18
[RHOSINFRA-148] Adds internal DNS to OpenStack's VMs

aopincar:


aopincar:
Implemented in PR #191

PR #175, opened 2016-04-18
[RHOSINFRA-104] - Add external swift node

ovorobio:


settings/installer/ospd/ospd.spec
```diff
@@ -53,6 +53,13 @@ subparsers:
                     help: The overcloud storage type
                     default: no-storage.yml
 
+            - title: Swift object-storage
```
tkammer: no need to make this a dedicated flag, use teh "storage" flag and make a "swift.yml" option there

playbooks/installer/ospd/main.yml
```diff
@@ -58,6 +58,8 @@
   roles:
       - {role: installer/ospd/overcloud/storage, when: "installer.storage is defined"}
       - {role: installer/ospd/overcloud/network, when: "installer.network is defined and installer.network.isolation.enable == 'yes'"}
+      - {role: installer/ospd/overcloud/swift, when: "installer.swift is defined"}
```
tkammer: There is no need for a dedicated role.
please move this under "installer/ospd/overcloud/storage", see how it does the configurations of different storage backends and add swift there

yfried-redhat:
LGTM. do we have a poc job?

PR #180, opened 2016-04-19
[RHOSINFRA-147] ir-ci jenkins POC 

obaranov:
Jenkins job execution from CLI

Usage example:

``` shell
# generate conf file
ir-provisioner virsh --generate-conf-file=job_params.ini
ir-installler ospd --generate-conf-file=job_params.ini
# edit generated file and provide all the required args there
# create jenkins config file 
cat jenkins.ini
[jenkins]
url = http://10.8.180.145:8080/
username = obaranov
password = mypass
job_name = poc-ir-obaranov-distro

# run jenkins job
ir-ci -d jenkins --config=jenkins.ini --from-file=job_params.ini
```

Job POC: http://10.8.180.145:8080/job/poc-ir-obaranov-distro/

Known issues and limitation: 
1) Optional arguments should be manually added to the job_params.ini or handled on job itslef (see the images-url usage in job configuration) 
2) We pass job parameters as the environment variables to the infrared. So I think we need to have more unique names for these variables to avoid duplication from installer and provisioner. E.g. we have have the same flags in different specs and in this case we will lose one of them...

### DIFF
--- a/cli/ci.py
+++ b/cli/ci.py
@@ -1,0 +1,62 @@
+import ssl
+import jenkins
+
+from cli import logger
+
+LOG = logger.LOG
+
+
+class JenkinsRunner(object):
+    """
+    The jenkins job runner.
+
+    Starts jenkins jobs using Jenkins python API.
+    """
+    START_TIMEOUT = 10
+    BUILD_COMPLETION_TIMEOUT = 60*60*3
+    START_PULL = 1
+
+    def __init__(self, auth_info, ignore_ssh_warnings=True):
+        self.host = auth_info.get('url')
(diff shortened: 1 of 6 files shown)


tkammer:
TBH, I feel that the "INI" approach was wrong and we should have created a YAML per our other settings (consistence).
If you move to YAML, you can simply put it under "provisioner:" or "installer" so no duplication problem :)
Also, I don't think we should support ENV variables, it's a source for problems as the user always "forgets" he set an ENV variable that mess everything up :)
If we already have the "--from-file" and CLI, lets remove ENV.

PR #183, opened 2016-04-20
[RHOSINFRA-154] Fix Babel in requirements.

obaranov:
@yfried-redhat, can you please review?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,5 @@
+# temporary workaround to fix version conflict issue
+Babel!=2.3.0,!=2.3.1,!=2.3.2,!=2.3.3,>=1.3
 ansible>=2.0.0
 clg>=2.0.0
 colorlog>=2.6.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,9 @@
-# temporary workaround to fix version conflict issue
-Babel!=2.3.0,!=2.3.1,!=2.3.2,!=2.3.3,>=1.3
-ansible>=2.0.0
+shade>=1.7.0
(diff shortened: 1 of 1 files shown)


tkammer:
The problem is specific for Fedora23, I think we should add a Doc comment that Fedora23 is currently problematic.

obaranov:
I think the problem is not only with Fedora 23. The problem is in the python-novaclient**<3** IR requirement. 

If installs 2.3.3 Babel. And this version is not supported by the recent python-heatclient package (heaclient has not restrictions)

yfried-redhat:
retest openstack

yfried-redhat:
retest virsh

requirements.txt
```diff
@@ -1,10 +1,9 @@
...
-ansible>=2.0.0
+shade>=1.7.0
+python-novaclient>=2.21.0,!=2.27.0,!=2.32.0
+ansible>=2.0.0,!=2.0.2
 clg>=2.0.0
 colorlog>=2.6.1
 ipaddress==1.0.16
-python-novaclient<3
 PyYAML>=3.11
 shade>=1.7.0
```
yfried-redhat: shade is here twice

yfried-redhat:
retest openstack

yfried-redhat:
retest openstack

yfried-redhat:
retest openstack

requirements.txt
```diff
@@ -1,10 +1,12 @@
-ansible>=2.0.0
+pip>=8.0.0
```
yfried-redhat: I don't think this means anything. AFAIK, the new pip will be active only after all packages were installed.
Let's try it without this line

PR #186, opened 2016-04-21
Get selinux binding during installation

yfried-redhat:
File modules (copy/file/template) require selinux binding, which isn't
provided in pypy and thus unsupported by virtualenv.
Get it from system during installation.

### DIFF
--- a/docs/source/setup.rst
+++ b/docs/source/setup.rst
@@ -33,12 +33,11 @@ to avoid corrupting the system packages::
   $ virtualenv .venv
   $ source .venv/bin/activate
 
-`Get python binding for SELinux <https://dmsimard.com/2016/01/08/selinux-python-virtualenv-chroot-and-ansible-dont-play-nice/>`_::
 
-  $ cp -rv /usr/lib64/python2.7/site-packages/selinux/ $VIRTUAL_ENV/lib/python2.7/site-packages
-
-
-.. note:: libselinux-python is in `Prerequisites`_ but doesn't have a pip package. Create virtualenv with site packages enabled to avoid this hack::
+.. note:: libselinux-python is in `Prerequisites`_ but doesn't have a pip package.
+ When detecting a virtualenv without selinux binding, InfraRed will try to
+ `get python binding from system <http://dmsimard.com/2016/01/08/selinux-python-virtualenv-chroot-and-ansible-dont-play-nice/>`_.
+ To avoid this, create an "open" virtualenv with site packages enabled::
 
   $ virtualenv --system-site-packages .venv
 
--- a/setup.py
+++ b/setup.py
@@ -58,9 +58,6 @@ except ImportError as e:
             "selinux")
         if not os.path.exists(SELINUX_PATH):
             raise new_error
-        print("missing selinux binding. Trying to pull modules from "
-                 "system ({source}) to virtualenv {dest}".format(
-            source=SELINUX_PATH, dest=sys.prefix))
 
         # filter precompiled files
         files = [f for f in os.listdir(SELINUX_PATH)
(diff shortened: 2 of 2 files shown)


bregman-arie:
Sorry, but I'm not sure I understand the issue.

" require selinux binding, which isn't provided in pypy and thus unsupported by virtualenv." ->
1. what selinux binding? 
2. what is the current error?
3. "pypy" -> you meant PyPI right?

bregman-arie:
I don't like the idea of adding those lines to every task, so -1.

I'm just thinking about potential contributors who are familiar with Ansible, adding new task and then get -1 each time because they didn't add those lines. Awful.

fhubik:
This issue is another blocker for RHEL-7.2.

I am not sure if "forcing" system ansible every time is best way, because:
a) IR developers might not be aware of this problem in future, they may forgot to force
b) Is not flexible forcing on Play/(even)Task level, the code will be mess in the end
c) This needlessly overcomplicate things

I would rather consider one-time venv "hack"  in the beggining of every major playbook (prov/inst/tester), e.g. this way (working for me):

$ cat ./playbooks/provision.yml
# Begin of the file
hosts: localhost
sudo: yes
- name: Install selinux bindings as they are required for file ops of ansible modules in venv
  yum: name=libselinux-python state=present
  when: in_venv

- name: Hack venv once and for all for this standalone major playbook (copy libselinux-python from system to current venv if defined)
  shell: |
      cp -R /usr/lib64/python2.7/site-packages/selinux  $VENV_DIR/lib64/python2.7/site-packages
  when: in_venv
...

bregman-arie:
Totally agree.

You are basically telling developers "in ansible you simply had to use file/copy/template as they thought you and showed you in the examples but, here, in IR, you always need to add:

 vars:
         ansible_python_interpreter: "{{ sys_python | default('/usr/bin/python') }}"

And there is probably no chance they will know as new contributors, until they bump into it by getting failures. So -2 on this change.

queria:
As point 1 in http://infrared.readthedocs.io/en/devel/quickstart.html is yum-installing libselinux-python, i would suggest to solve it by using open venv (with system site packages), as that can also save some bandwitdh/time in jenkins jobs when re-creating the venvs etc.

yfried-redhat:
@fhubik, Your way is actually better. I'll look into doing it (either in ansible or in setup script somehow)
@queria, We've tried working with open venv. It's listed as an alternative in docs, but you get all other packages from system as well and it gets problematic with version conflicts.

Best solution would be to work in system mode as default. I've heard some suggestions about creating rpms in PyconIL but I currently don't have the time to investigate.

queria:
@yfried-redhat
1) for clients/libs where we do need specific newer/older version we should be explicit in requirements (and/or possibly allow updates)
2) system mode is terrible as it just collides with rpm systems (or any other pkg management), and me for one but not only, are never gonna let 3rd party tools touch system libs (neither i would advice to anyone to play Russian roulette with their system), also it does not have solution for how to isolate from other tools (e.g. old/stable infrared, khaleesi, other openstack clients or ansible based tools etc), it can though work in controlled one purpose cases like jenkins-slave-just-for-infrared-jobs ... though even there you can hit issues, like what to have in system if you need to test patch upgrading requirements (or ansible or ...) ... based on all this, it seems the (open?)venv usecase will need to be (supported/)possible in some way (as secret knowledge of black magic maybe, if just single users will do it on their own) so maybe it would be better to have it as also the main way ...

though it of course depends, maybe the issues with incorrect clients version etc are too deep so that price of dealing with them would be too high? is that the case?

fhubik:
+1 Looks god to me as this is gold way between hacking every playbook at the beginning and modifying python interpreter every time when "file" related module is used.

queria:
Anyway I agree that this patch can serve as good solution, way better then having extra python interpreter, or any other, overrides spread throughout the yamls.
Also for open venv's with libselinux-python available it's almost noop, so +1 for me too.

yfried-redhat:
Verified by poc job https://rhos-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/poc_yfried_venv/49/

yfried-redhat:
retest openstack

yfried-redhat:
retest openstack

PR #187, opened 2016-04-21
Add tester

bregman-arie:
Before this change, the only node in tester group
was undercloud.

There is a use case ( component testing ) for testing
on one provisioned node and this is way 'tester' node
was added to the topology.

### DIFF
--- a/settings/provisioner/topology/tester.yml
+++ b/settings/provisioner/topology/tester.yml
@@ -1,0 +1,25 @@
+name: "tester"
+amount: "1"
+cpu: "8"
+memory: "8192"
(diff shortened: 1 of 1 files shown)


tkammer:
The "tester" group shouldn't be inside the topology, it should be a flag in "ir-tester" that will add the "Tester" group to a node
something like: "ir-tester tempest --host=controller"

yfried-redhat:
@tkammer What about an external tester node?

yfried-redhat:
I think we need both -  Users will choose if they want a tester node reuse an existing node

PR #188, opened 2016-04-21
[RHOSINFRA-158] Spec improvements. Integration tests

obaranov:
Example: 
```
[ospd]
# the installation user password
# default value: stack
user-password = 123123
# the installation user name
user-name = stack
# the network protocol for overcloud
# default value: ipv4
network-protocol = ipv4
# specifies whether ths ssl should be used for overcloud
# default value: no
overcloud-ssl = no
# the overcloud network backend.
# default value: vxlan
network-backend = vxlan
# the firewall configuration
# default value: default.yml
# allowed values: { default.yml }
firewall = default.yml
# the overcloud storage type
# default value: no-storage.yml
# allowed values: { ceph.yml, netapp-iscsi.yml, no-storage.yml, netapp-nfs.yml, ceph-external.yml }
storage = no-storage.yml
# the product core build
# default value: latest
product-core-build = latest
# the product build
product-build = latest
# the undercloud config details
undercloud-config = default.yml
# the product core version
# allowed values: { 7, 8 }
product-core-version = Edit with one of ['7', '8'] options, OR override with CLI: --product-core-version=<option>
# specifies whether the images should be built or imported
# allowed values: { import, build }
images-task = Edit with one of ['import', 'build'] options, OR override with CLI: --images-task=<option>
# the overcloud network isolation type
# allowed values: { no-isolation.yml, three-nics-vlans.yml }
network-isolation = 12312
# the product version
product-version = Edit with one of ['7', '8'] options, OR override with CLI: --product-version=<option>


```


cli/spec.py
```diff
@@ -549,50 +551,87 @@ def _generate_config_file(
...
+                out_config.set(
+                    subcommand,
+                    opt,
+                    out_config_old.get(subcommand, opt))
 
     with open(file_name, 'w') as configfile:  # save
         out_config.write(configfile)
 
 
+def _get_short_help(option_attributes):
```
yfried-redhat: Why short? let's get the full help

obaranov: done

cli/main.py
```diff
@@ -130,7 +124,7 @@ def get_settings_files(self):
         settings_files = []
 
         # first take all input files from args
-        for input_file in self.args['input'] or []:
+        for input_file in self.args.get('input', []) or []:
```
yfried-redhat: Either `args.get('input') or []` or `args.get('input', [])`

Do we expect `args["input"]` to hold something other than a list?

cli/spec.py
```diff
@@ -162,7 +162,7 @@ def get_file_locations(cls, settings_dir, subcommand, arg_name):
 
     @classmethod
     def get_allowed_files(cls, settings_dir, subcommand, arg_name,
-                          search_root=False):
+                          search_root=False, file_extensions=('yml', 'yaml')):
```
yfried-redhat: nice

cli/main.py
obaranov: We need that for the case when 'input' argument it not in spec. Without that change we will get an exception saying the 'input' key is not present in input.

obaranov: When 'input' is present in spec file but not provieded via CLI, argparse will assign it to None. 
So self.args['input'] == None.

yfried-redhat:
retest openstack

cli/main.py
yfried-redhat: OK

yfried-redhat:
Updating to latest master because gate fails

yfried-redhat:
retest openstack

PR #193, opened 2016-04-26
Change topology format

mpryc:
This proposes new topology format. Instead of using:
topology=N_something, Y_something2, Z_something3

The change proposes:
topology=something:N, something2:Y, something3:Z


rhosqeauto:

rhosqeauto:

bregman-arie:
Looks good. Is there a poc job that use this change?

mpryc:
@bregman-arie no, there is no poc job but I've tested the change in the small python code so no regression is introduced and string is properly parsed.

yfried-redhat:
test this please

rhosqeauto:

mpryc:
The change was tested on the migi-poc-gate-InfraRed-ospd_qe-openstack-pull-193 job with success.

arxcruz:
Is the old X_something still works? If we introduce this change, all our jobs based on X_something will start to fail, I would like to see compatibility, and set in the documentation that X_something will be deprecated soon.

bregman-arie:
@arxcruz since we have stable version for IR, I think it should be fine merging this to master. Jobs should use stable version.

yfried-redhat:
ok to test

cli/spec.py
```diff
@@ -222,11 +222,21 @@ def resolve_value(self, arg_name, defaults=None):
...
+
+            node_type, number = None, None
+
+            pattern = re.compile("^[A-Za-z]+:[0-9]+$")
             if pattern.match(topology_item) is None:
-                raise exceptions.IRWrongTopologyFormat(self.value)
+                pattern = re.compile("^[0-9]+_[A-Za-z]+$")
+                if pattern.match(topology_item) is None:
+                    raise exceptions.IRWrongTopologyFormat(self.value)
+                number, node_type = topology_item.split('_')
```
yfried-redhat:
```python
LOG.warn("This topology format is deprecated and will be removed in future versions. Please see `--help` for proper format") 
```

mpryc: added as LOG.warning

arxcruz:
@bregman-arie not exactly, we have nightly jobs running on master, besides, when master became stable, we need to change all JJB, it can easily broken our gate jobs.

mpryc:
@arxcruz the last fix introduced backwards compability so nothing will be broken.

arxcruz:
@mpryc Awesome! Thanks!

yfried-redhat:
@mpryc Thank you. merged.

PR #201, opened 2016-05-03
[RHOSINFRA-142] Creates specific SPEC files with inventory to apps

aopincar:
Creates specific SPEC file for each app.
Moves the inventory option from the base's spec to the app's spec.

Resolves RHOSINFRA-142


aopincar:
retest openstack

aopincar:
retest virsh

PR #217, opened 2016-05-09
Add "Fix addition / creation of authorized_keys" patch from master

fhubik:
We need to fix this asap as deployment is failing for users
with non-prepared slaves (like cleanly reprovisioned nodes).
Our slaves doesn't have this problem, this file exists there.

Part of Arx's PR proposed here:
https://github.com/rhosqeauto/InfraRed/pull/215/commits/8c87edacfddd59e27c511533de42c71666db7ab2

### DIFF
--- a/playbooks/installer/ospd/virthost.yml
+++ b/playbooks/installer/ospd/virthost.yml
@@ -116,5 +116,7 @@
         lineinfile:
             dest: "/home/{{ installer.user.name }}/.ssh/authorized_keys"
             line: "{{ undercloud_public_key.stdout }}"
+            create: yes
+            mode: o-rwx,g-rwx,u+rw
         become: yes
         become_user: "{{ installer.user.name }}"


playbooks/installer/ospd/virthost.yml
```diff
@@ -116,5 +116,7 @@
         lineinfile:
             dest: "/home/{{ installer.user.name }}/.ssh/authorized_keys"
             line: "{{ undercloud_public_key.stdout }}"
+            create: yes
```
obaranov: I would recommend to use the athorized_keys module instead of lineinfile here: 

```
       authorized_key:
           user: "{{ installer.user.name }}"
           key: "{{ undercloud_public_key.stdout }}"
```

This module will create file if it is not present

arxcruz: Didn't knew about this module, I think it's better!

fhubik:
Ok, we have private poc job confirmed this will fix stable branch.

arxcruz:
@fhubik Which fix? The one using authorized_key or lineinfile?

fhubik:
The one using lineinfile.

tkammer:
@fhubik please make sure it works with "authorized_keys" before merging, we should strive to use Ansible modules where possible.

yfried-redhat:
retest openstack

playbooks/installer/ospd/virthost.yml
yfried-redhat: Please use @obaranov's suggestion, and verify that it works

yfried-redhat:
retest openstack

yfried-redhat:
retest openstack

yfried-redhat:
This needs to be rebased in order for selinux fix to work

yfried-redhat:
stable was updated with required changes

PR #221, opened 2016-05-12
Adding openstack logs to collect-logs

arxcruz:
This will collect logs from openstack. Based on http://docs.openstack.org/openstack-ops/content/logging_monitoring.html

### DIFF
--- a/playbooks/collect-logs.yml
+++ b/playbooks/collect-logs.yml
@@ -34,6 +34,10 @@
       file: dest=/var/log/extra state=directory
       ignore_errors: true
 
+    - name: Prepare directory with openstack logs
+      file: dest=/var/log/extra/openstack state=directory
+      ignore_errors: true
+
     - name: collect logs from all failed systemd services
       shell: |
         systemctl -t service --failed --no-legend | awk '{print $1}' \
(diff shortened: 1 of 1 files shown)


playbooks/collect-logs.yml
```diff
@@ -114,6 +118,36 @@
...
+    - name: Collect logs from neutron
+      sudo: yes
+      shell: cp -Rf /var/log/neutron/* /var/log/extra/openstack
+      ignore_errors: yes
+
+    - name: Collect logs from horizon
+      sudo: yes
+      shell: cp -Rf /var/log/httpd/horizon_error.log /var/log/extra/openstack
+      ignore_errors: yes
+
```
obaranov: Just one suggestion. Maybe replace all these task with one tasks with the with_items or with_dict (up to you)?

arxcruz: I though about that, but, what if in the middle, it doesn't find for example the glance files, the with_items will continue with the others or it will fail and move to next task?

obaranov: with_items will run over all the items even if someone will fail. If you put ignore_errors: yes it will ignore any errors. I've tested that with simple playbook: 
```
- hosts: localhost
  tasks:
    - file:
        path: "{{item}}"
        state: file
      ignore_errors: yes
      with_items:
        - /bin/true
        - "I am here"
        - /bin/false
```

yfried-redhat: Collect-logs is based on grabbing everything in `{{ job.archive }}` list see [link](https://github.com/rhosqeauto/InfraRed/blob/a3bbb4448bade115b08aaa8ddc2dc566fe13e5d8/playbooks/collect-logs.yml#L132-L140). 

Please according to [`ospd.yml`](https://github.com/rhosqeauto/InfraRed/blob/9471828d8bf4088f9920216650afb996d733f39f/settings/installer/ospd/ospd.yml#L13-L13) the entire `/var/log/*` tree should be archived in all nodes.

Please investigate why this doesn't apply for undercloud node

arxcruz:
Obsolete

PR #224, opened 2016-05-14
Create universal role and job for automated build/upload for OC images

fhubik:
Also, completely rewrite the old one, make this role
work with CI build and/or individual one (during OSPd deployment).

### DIFF
--- a/playbooks/build-oc-images.yml
+++ b/playbooks/build-oc-images.yml
@@ -1,5 +1,11 @@
 # This playbook is not included in "ir-XYZ" flows as is not
 # related to any of them and is used only stand-alone.
+# 
+# This playbook is designed to run on single node in virtual
+# machine environment. Currently node called "controller"
+# is used as main node doing this task as IR doesn't support
+# creating stand-alone single node topology and we still
+# need to have this playbook integrated with current IR workflow.
 ---
 - name: Store local public key as variable for current user
   hosts: controller
--- a/playbooks/build-oc-images.yml
+++ b/playbooks/build-oc-images.yml
@@ -1,3 +1,5 @@
+# This playbook is not included in "ir-XYZ" flows as is not
+# related to any of them and is used only stand-alone.
 ---
 - name: Store local public key as variable for current user
   hosts: controller
--- a/playbooks/build-oc-images.yml
+++ b/playbooks/build-oc-images.yml
@@ -44,6 +44,7 @@
       task: 'build'
       ospd_version: "{{ ospd_version }}"
       base_image_url: "{{ private.distro.rhel.guest_image_url }}"
+      upload_to: 'remote'
       storage_user: "{{ private.storage.oc_image_repo.user }}"
       storage_server: "{{ private.storage.oc_image_repo.server }}"
       storage_path: "{{ private.storage.oc_image_repo.path }}"
--- a/playbooks/build-oc-images.yml
+++ b/playbooks/build-oc-images.yml
@@ -1,0 +1,51 @@
+---
+- name: Store local public key as variable for current user
+  hosts: controller
+  gather_facts: no
+  tasks:
+    - name: Create and save public key from private key
+      shell: "ssh-keygen -y -f {{ hostvars[inventory_hostname].ansible_ssh_private_key_file }}"
+      register: cmd_priv_key
+      delegate_to: localhost
+
+- name: Enable root access as cloud images might not have one first
+  hosts: controller
+  gather_facts: no
+  become: yes
+  become_user: root
+  tasks:
+    - name: Add local public key to root's authorized_keys
+      authorized_key:
+        key: "{{ cmd_priv_key.stdout }}"
+        user: root
+
+    - name: Copy local private key to remote root
+      copy:
+          src: "{{ hostvars[inventory_hostname].ansible_ssh_private_key_file }}"
+          dest: "/root/.ssh/id_rsa"
+
+    - name: Copy local public key to remote root
+      copy:
+          content: "{{ cmd_priv_key.stdout }}"
+          dest: "/root/.ssh/id_rsa.pub"
+
+    - name: Fix permissions for both keys
+      shell: |
+            chmod 750 /root/.ssh/id_rsa
+            chmod 700 /root/.ssh/id_rsa
+
+- name: Build images as root and upload them to remote storage
+  hosts: controller
+  gather_facts: yes
+  become: yes
+  become_user: root
+  roles:
+    - role: installer/ospd/images
+      task: 'build'
+      ospd_version: "{{ ospd_version }}"
+      base_image_url: "{{ private.distro.rhel.guest_image_url }}"
+      storage_user: "{{ private.storage.oc_image_repo.user }}"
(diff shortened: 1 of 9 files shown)


playbooks/build-oc-images.yml
```diff
@@ -0,0 +1,51 @@
+---
```
yfried-redhat: Please add a comment here (and in commit message) explaining that this playbook is not exposed via IR interface (ie `ir-installer`) but is rather intended for advance usage via `ansible-playbook` command

fhubik: I believe this playbook doesn't fit to provision/install workflows and should be used only for building images, not actual deployments - therefore it shouldn't be included in CLI utils.

playbooks/installer/ospd/main.yml
```diff
@@ -19,9 +19,17 @@
   hosts: undercloud
   gather_facts: no
   become: yes
-  become_user: "{{ installer.user.name }}"
+  become_method: sudo
   roles:
-      - {role: installer/ospd/images}
+      - role: installer/ospd/images
+        task: "{{ installer.images.task }}"
+        work_path: "{%- if installer.images.task == 'import' %}{{ '/home/{{ installer.user.name }}' }}{% endif %}"
```
yfried-redhat: else?

playbooks/installer/ospd/main.yml
```diff
@@ -19,9 +19,17 @@
...
+        ospd_version: "{{ installer.product.core.version }}"
```
yfried-redhat: this is actually core (packstack version) so either you need `installer.product.version` or you need to change the name to `core_version`

roles/installer/ospd/images/defaults/main.yml
```diff
@@ -1,10 +1,17 @@
...
 
-tar_images: "{{ installer.images.files | default(tar_files[installer.product.version]) }}"
+# Mandatory role parameters
+base_image_url: 'UNDEFINED'
+rhos_release_rpm_link: 'UNDEFINED'
+upload_to: 'none'
+storage_server: 'UNDEFINED'
+storage_path: 'UNDEFINED'
+storage_ssh_key: 'UNDEFINED'
+storage_user: 'UNDEFINED'
```
yfried-redhat: No need to list mandatory args.
This defines defaults values as str("UNDFINED"/"none"). What we want is for ansible to throw errors in case of missing input

fhubik:
retest openstack

fhubik:
retest openstack

playbooks/installer/ospd/main.yml
fhubik: Else work_path will be taken from role defaults, as designed.

roles/installer/ospd/images/defaults/main.yml
fhubik: User may not have all mandatory params of certain role picked up from role's code, this way we can emphasize them a make role interface more clear - specifying mandatory params in defaults. Otherwise, without proper reading of code (and interpreting e.g. includes/conditionals) he may not notice at first look at "missing 'none'" whats wrong where. IMHO this really helps role interoperability.

playbooks/installer/ospd/main.yml
fhubik: Right, good point, I will fix in next patchset.

fhubik:
Rebased.

fhubik:
Poc's:

krcmarik:
Hi,

I am experiencing that my jobs are repeatedly failing on images download from slow storage share (bos), one solution could be to raise a timeout for image download in roles/installer/ospd/images/tasks/import.yml but If this pull request rewrites the functionality and make it possible to have own built images on faster storage share then It would be great to get this merged asap. Thanks

playbooks/installer/ospd/main.yml
yfried-redhat: Not sure. IIUC, else workpath="". I think what you need here is ternary operator:

```yaml
work_path: {{ (installer.images.task == 'import') | ternary( '~' |expanduser, default(omit) ) }}
```

krcmarik:
One more reason, It has been confirmed that the prebuilt images we are using are out of date and that's why we are not testing latest packages. The right way as It was said is to either use the prebuilt image and update them or maybe to build images agaisnt latest repo.

playbooks/installer/ospd/main.yml
```diff
@@ -19,9 +19,17 @@
   hosts: undercloud
   gather_facts: no
   become: yes
-  become_user: "{{ installer.user.name }}"
+  become_method: sudo
```
tkammer: if you have "become: yes" it defaults to "sudo" so no need for the "become_method: sudo" line

playbooks/build-oc-images.yml
```diff
@@ -0,0 +1,60 @@
...
+            chmod 700 /root/.ssh/id_rsa
+
+- name: Build images as root and upload them to remote storage
+  hosts: controller
+  gather_facts: yes
+  become: yes
+  become_user: root
+  roles:
+    - role: installer/ospd/images
+      task: 'build'
```
tkammer: instead of sending the params to the role, lets have everything in the role itself so a user can simply open the role and know what is being sent and from where.

tkammer:
Overall looks great, but I would like to talk about this as I have some concerns looking forward + some new requirements that were asked from us.
ping me on IRC when you're ready.

playbooks/build-oc-images.yml
fhubik: If we specify our playbook-related variables an params directly in role, we would destroy role reusability completely and would bemaking mess for anyone who wants to work with the role independently. We would end-up in with variables like "private.something.something2" directly in role again, like in khaleesi.
Instead, I think is better to have default (and empty "mandatory") params speficied in role context (there are files "defaults/main.yml" and "vars/main.yml") and we are overriding them with our playbook-related/private variables on this level, this way we are not polluting role code with unrelated/private/external variables.

playbooks/installer/ospd/main.yml
fhubik: Ah, ok ;)

playbooks/installer/ospd/main.yml
fhubik: Ah, thanks, missed this, will fix in next commit.

playbooks/build-oc-images.yml
mpavlase: Why should be default values inside role pre-filled by private.somekey values? It's definitely not a good idea.
If I want to use something, that is designed to be reusable, it's my own business to pass proper values as I need to get desired behavior. It doesn't matter if it's me as 'user', or 'author' of this file. The author is kind of 'user' as well.
Role is like a 'function' in any ordinary language with ability to define arguments. If there would be set some default values, it should be chosen very carefully and not force user to redefine them to 'empty values to get basic functionality.

In short, I definitely agree with @fhubik and vote +1 for implementation in this patch.

fhubik:
This PR is outdated and was moved to gerrithub for further work (https://review.gerrithub.io/#/c/278558/).

PR #235, opened 2016-05-19
[OSPD] Adds support to fetch images from RPM

yfried-redhat:
Started in OSPD 9 (and backported to 8), overcloud images are packaged
with product RPM.

Also requires some casting of arguments from int values to string.


yfried-redhat:
Tested with https://rhos-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Nightly/job/ospd_qe-nightly-9_director-puddle-rhel-7.2-virthost-3cont_1comp_1ceph-three-nics-vlans-ceph_internal-ipv4-vxlan-no-ssl

PR #236, opened 2016-05-19
Fix memory collect logs

arxcruz:
According ansible documentation, we should avoid use sudo when use fetch
module because it uses slurp to make the checksum and might consume all
the memory

### DIFF
--- a/playbooks/collect-logs.yml
+++ b/playbooks/collect-logs.yml
@@ -159,7 +159,6 @@
       ignore_errors: true
       when: job.gzip_logs is not defined
       sudo: no
-      validate_checksum: no
 
     - name: fetch log archive (tar)
       fetch: src=/tmp/{{ inventory_hostname }}.tar flat=yes dest={{ base_dir }}/{{ inventory_hostname }}.tar validate_checksum=no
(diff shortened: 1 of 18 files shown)


playbooks/collect-logs.yml
```diff
@@ -158,11 +158,13 @@
       fetch: src=/tmp/{{ inventory_hostname }}.tar.gz flat=yes dest={{ base_dir }}/{{ inventory_hostname }}.tar.gz validate_checksum=no
       ignore_errors: true
       when: job.gzip_logs is not defined
+      sudo: no
```
tkammer: please use "become: no" since "sudo" is deprecated

playbooks/collect-logs.yml
```diff
@@ -158,11 +158,13 @@
...
 
     - name: fetch log archive (tar)
       fetch: src=/tmp/{{ inventory_hostname }}.tar flat=yes dest={{ base_dir }}/{{ inventory_hostname }}.tar validate_checksum=no
       ignore_errors: true
       when: job.gzip_logs is defined and job.gzip_logs
+      sudo: no
```
tkammer: same as above

rhosqeauto:
ltgm, no need to sudo/become anyone, as these files should be readable anyway (mabe ensure that? chmod +r whatever.tar.[g/x/b]z )

queria:
ltgm, no need to sudo/become anyone, as these files should be readable anyway (mabe ensure that? chmod +r whatever.tar.[g/x/b]z )

PR #240, opened 2016-05-20
[RHOSINFRA-151] Added loadbalancer to Infrared

ovorobio:


yfried-redhat:
retest openstack

playbooks/installer/ospd/main.yml
```diff
@@ -2,6 +2,12 @@
 - include: pre.yml
   tags: backup
 
+- name: Deploy and configure external loadbalancer instance
```
tkammer: Please move this into the loadbalancer invocation below.
You can read this reference as to what I mean.
http://docs.ansible.com/ansible/playbooks_roles.html#role-dependencies

Issue #229, opened 2016-05-17
overcloud VMs should use the same naming as the they get deployed by the ospd installer


mcornea:
Currently Infrared creates the VMs with an index starting at 1 if there are multiple nodes of the same type. For instance:

Id    Name                           State
---------------------------------------------------
17    undercloud                   running
25    controller1                    running
28    compute                       running
29    controller2                    running
30    controller3                    running

The installer deploys the nodes by always having an index starting at 0:

+-------------------------+-
| Name                        | 
+-------------------------+-
| overcloud-compute-0     | 
| overcloud-controller-0  | 
| overcloud-controller-1  | 
| overcloud-controller-2  | 
+-------------------------+-

It would be great to have Infrared create the VMs used for overcloud nodes with the same naming - always use an index starting at 0, e.g:

 Id    Name                           State
---------------------------------------------------
17    undercloud                   running
25    controller-0                    running
28    compute-0                       running
29    controller-1                    running
30    controller-2                    running

yfried-redhat:
The 0 based issue was fixed.
I think this is addressed by RHOSINFRA-194.
If not, please open a JIRA item or comment there

PR #241, opened 2016-05-20
Change overcloud VMs name to start at index 0

mcornea:
This change enables the creation of overcloud VMs by using an index starting at 0 in the form of <role>-<index>

It addresses issue #229


rhosqeauto:

rhosqeauto:

tkammer:
Thanks for the help!
We're currently switching our bash to ansible tasks, would you mind rebasing this on the newly supported playbook?

yfried-redhat:
In #244 the the way we create and name nodes is being refactored. Could you please review and make sure it suits your request?

yfried-redhat:
This is partially addressed by #244. Closing.

PR #247, opened 2016-05-23
Enable ansible-lint for the InfraRed project

mpryc:
This change enables ansible-lint for the playbooks in the
InfraRed project.

Two of the rules are disabled and they will be enabled
once all playbooks will change to allow pass those:

    ANSIBLE0006,ANSIBLE0007

Closes: #246

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ script:
     - pep8 cli
     - pep8 tests
     - py.test tests -v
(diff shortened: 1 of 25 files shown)


mpryc:
I am unsure if !placeholder can be surrounded with the quotes. Without this ansible-lint throws error. Other way around it is to use escaping: \!placeholder

PR #253, opened 2016-05-25
[RHOSINFRA-181] Allows to set custom hostnames for controllers

ovorobio:


roles/installer/ospd/overcloud/tagging/tasks/main.yml
```diff
@@ -33,3 +33,8 @@
   with_nested:
       - "{{ node_list.results }}"
       - "{{ tagged_flavors }}"
+
+- name: update capabilities for controllers
+  shell: "source ~/stackrc; ironic node-update {{ item.1 }} replace properties/capabilities='node:controller-{{ item.0 }},boot_option:local'"
+  when: "installer.ovbercloud.hostname.controller is defined and installer.overcloud.hostname.controller == 'yes'"
```
tkammer: note the type of "overcloud"

settings/installer/ospd/ospd.spec
```diff
@@ -45,6 +45,9 @@ subparsers:
                     type: Value
                     help: Specifies whether ths SSL should be used for overcloud
                     default: 'no'
+                overcloud-hostname:
```
tkammer: I think that "features" might need to move to a different structure.. I mean, more and more features are going to be added so I think we should have a distinction between a "deployment" need (such as building a LB for example) and a feature of OSP-d, which might "pollute" our .spec file.
I think we need to think about this further, maybe moving it to a configuration file? (thinking out loud)

ovorobio: Yes, it can be separated to different folder, we can use the feature we already have to pull settings from different folders and create setting folders for different roles/features: controllers, computes, nova, neutron or something like that.

roles/installer/ospd/overcloud/hostname/templates/hostnames.yml.j2
```diff
@@ -0,0 +1 @@
+{{ installer.overcloud.hostname.template.content |to_nice_yaml }}
```
yfried-redhat: why do we need a whole file to hold a single line? can't this be done in-line in playbook file?

roles/installer/ospd/overcloud/tagging/tasks/main.yml
```diff
@@ -33,3 +33,12 @@
   with_nested:
       - "{{ node_list.results }}"
       - "{{ tagged_flavors }}"
+
```
tkammer: why can't this be part of the role for hostnames? (as preparation)
I would prefer not to "pollute" a general playbook with "task specific" stuff.

settings/installer/ospd/overcloud/hostname/controllers.yml
```diff
@@ -0,0 +1,10 @@
+controller: 'yes'
```
tkammer: Same here, I think this should be part of the `role` rather than a user set param.
If a user wants to invoke this feature, it's a 'static' template that should reside under the role itself

tkammer:
:+1: once gate pass, please merge

tkammer:
@yfried-redhat  ^

ovorobio:
retest openstack

ovorobio:
retest openstack

PR #255, opened 2016-05-26
Avoid files bigger than 1GB be added in the collect logs tarball

arxcruz:
This is necessary in case we are building images and since these images
have usually more than 10GB, when we collect the logs, the tarball
became huge unecessary.

### DIFF
--- a/playbooks/collect-logs.yml
+++ b/playbooks/collect-logs.yml
@@ -136,7 +136,7 @@
     - name: collect logs
       shell: |
         mkdir -p /tmp/{{ inventory_hostname }};
-        find {{ job.archive|join(' ') }} -maxdepth 2 -type f -size -1024M -not -path '*/\.*' -exec cp -rL --parents {} /tmp/{{ inventory_hostname }} \;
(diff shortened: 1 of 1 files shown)


playbooks/collect-logs.yml
```diff
@@ -136,9 +136,7 @@
     - name: collect logs
       shell: |
         mkdir -p /tmp/{{ inventory_hostname }};
-        for F in $(ls -d1 /var/log/rpm.list /var/log/extra {{ job.archive|join(' ') }}); do
-          cp -rL --parents $F /tmp/{{ inventory_hostname }}
-        done;
+        find {{ job.archive|join(' ') }} -maxdepth 2 -type f -size -1024M -not -path '*/\.*' -exec cp -rL --parents {} /tmp/{{ inventory_hostname }} \;
```
yfried-redhat: Isn't this ignoring `/varlog/rpm.list` and `/var/log/extra` that were mentioned previously?

arxcruz: You're right. Updating.

Issue #256, opened 2016-05-26
Ability to deploy not "latest" puddle broken


fhubik:
We don't have working ability to deploy with non-latest puddle of certain OSPd version - if the last one is broken, users can not do it directly.

fhubik:
Fix is result of discussion at https://github.com/rhosqeauto/InfraRed/issues/171, proposed at https://github.com/rhosqeauto/InfraRed/pull/257.

Issue #171, opened 2016-04-17
confusing 'ir-installer ospd' arguments


bregman-arie:
I wanted to run 'ir-installer ospd' on my openstack-ovb environment, but the required arguments made me confused. These are the required arguments:

product-core-version
product-version
images-task
network-isolation

Two questions/requests:

1. What is difference between product-version and product-core-version? I see it can receive the same values but it's not clear from the description what is the difference between the two. could you please extend the description? also, is it possible to determine the value of one them based on the second argument value?

2. network-isolation description is "The overcloud network isolation type" - I really have no idea what types exist. can you extend the description to include some examples for types? also, does it must to be required? can a default be set?

yfried-redhat:
1. `product-version` applies to ospd version. core bits (packstack, nova, neutron, etc...) **of the _undercloud_** will match the default unless specified in `prodcut-core-version`. Overcloud version is currently decided by the default images the ospd downloads, so it's hard-coupled to the ospd version ATM. This is standard `rhos-release` usagae
2. @tkammer @obaranov could you please answer?

bregman-arie:
@yfried-redhat sorry but not sure I'm following. 

"product-version applies to ospd version." ->  by ospd you mean undercloud? overcloud? 

"core bits (packstack, nova, neutron, etc...) of the undercloud will match the default unless specified in prodcut-core-version" -> so undercloud version is set by 'product-core-version'?

why isn't there one option to specify version for everything?

Can someone please add the necessary documentation. this is quite confusing...

tkammer:
The reason there is no "one flag" for both versions is because from OSP8 we need to support undercloud version 8 and overcloud with 7. (basically install OSP7 overcloud using OSP8 undercloud)
Because of this, we made this specific.
But I do agree that we need to make our documentation better..

bregman-arie:
Thanks for the answers. 

In my opinion this issue should be closed after:

1. 'no-isolation' set as the default value for network-isolation
2. better documentation on 'product-version', 'product-core-version' and the relationship between them.

*3. should a default be set for image-task also?

fhubik:
Ad 2. - I was also confused about these params as it is not documented, simply:
  --product-version {7,8}
                        The product version
                        Allowed values: { 7, 8 }
  --product-version {7,8}
                        The product version
                        Allowed values: { 7, 8 }
is not enough.

fhubik:
Luigi pointed out to another related weird code here. We have in several playbooks code:
```
[...]

PR #257, opened 2016-05-26
Fix broken handling of rhos-release

fhubik:
Rhos-release tool is not maintaining dependencies
between ospd and core versions properly when "latest"
is not specified. Therefore ability to specify certain
version numbers must be working (until this broken) and called
in specific order. First, we need to call rhos-release with
{{ installer.product.version }}-director and version number
and then {{ installer.product.core.version }} with different
specific core version number. Otherwise, rhos-release end up
in creating repositories with links to non-existing repos
and deployment will fail.

This fixes issues 171 and 256.

### DIFF
--- a/playbooks/installer/ospd/virthost.yml
+++ b/playbooks/installer/ospd/virthost.yml
@@ -31,10 +31,10 @@
         delay: 5
 
       - debug:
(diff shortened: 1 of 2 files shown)


fhubik:
retest openstack

yfried-redhat:
Fix #171 
Fix #256

PR #258, opened 2016-05-26
[RHOSINFRA-180] added reboot handler to ospd installer

ovorobio:


tkammer:
:-1: This creates a dependency between the installer and the virsh provisioner and breaks the goal of being idempotent.
Lets talk about this

PR #260, opened 2016-05-30
Index VMs from virsh from 0 (previousely 1)

mpavlase:


yfried-redhat:
@mpavlase Looking at the gate, I don't see any VMs created on the virthost

roles/provisioner/virsh/tasks/create_vms.yml
```diff
@@ -82,7 +82,7 @@
          --noautoconsole \
          --autostart \
          --vnc
-  with_sequence: count={{node_data.amount | int }}
+  with_sequence: start=0 count={{node_data.amount | int - 1 }}
```
yfried-redhat: please apply this for `openstack` roles as well, as discussed.

mpavlase: ok

PR #266, opened 2016-06-17
Correct usage of rhos-release to always pin

yazug:
include -P to pin all versions installed and not point at latest
This prevents the repo from shifting if a new version comes out
while a test is underway

Change-Id: I699092949f09c16247f7e7f032dd0afa3fab40c3


yazug:
moved to a gerrit review - https://review.gerrithub.io/#/c/280853

PR #279, opened 2016-12-21
fixes #278 : enablement of Travis CI builds

### LINKED ISSUE #278: enable Travis CI on infrared
Please enable Travis CI on infrared repository, this is an essential tool for providing feedback regarding infrared stability and also helps validating PRs.

I already did the work needed for enabling Travis in my working fork and you can see the results at https://travis-ci.org/ssbarnea/InfraRed/branches 

Current Travis configuration is to perform 3 builds: 
* py27-linux(all tests)
* py35-linux
* py27-osx

This should catch most errors without the risk of creating too many builds. 

Please note that I am willing to take care of Travis integration if further help is needed. 

I know that we already have internal gates for testing it but using Travis in parallel is much more open approach because anyone forking infrared can use travis to test his changes, something that is not possible with gerrit approach.

I will open both a PR on github and a Gerrit review with these changes.

ssbarnea:
Please excuse the size of this change but this was the only way to enable Travis while testing for py2 and py3. 

The current code of infrared is broken on py3 due to few bugs fixed by this change. I guess at this moment there is no gate for testing py3, but this patch enables one.

One change that may come bit weird is the upgrade of ansible to 2.2, that's needed because only Ansible 2.2 is py3 compatible.

Same applies to ansible-lint which was a very old version, upgrading it allowed us to run py3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
   - "2.7"
@@ -6,7 +7,8 @@ python:
 before_script:
   - export PATH=$HOME/.local/bin:$PATH
 install:
-- pip install --upgrade pip tox-travis
+# if setuptools<17.1 installing even dependencies could fail
+- pip install --upgrade pip setuptools tox-travis
 - pip install .
 # command to run tests
 script:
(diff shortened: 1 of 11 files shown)


leifmadsen:
Resolves #278

PR #282, opened 2017-04-04
Fixing a type in example

agurenko:
memeory -> memory


ssbarnea:
Please use git-review to contribute changes. We do not use GitHub PRs.

PR #298, opened 2017-08-26
Add tags to network related virsh cleanup tasks

dprince:
This patch tags the tasks in the virsh/cleanup playbook
relate to network cleanup with 'networks'. This is useful
if you want to skip infrared's network configuration by
using Ansible's skip-tags option on the CLI.

Example:

infrared virsh --host-address 127.0.0.1 --host-key ~/.ssh/infrared \
--cleanup yes --ansible-args="skip-tags=networks"

The motivation here is I'm using infrared to provision an undercloud
VM but I have my own networking setup (custom bridges, switches, etc)
for baremetal and would prefer to manage it on my own.

### DIFF
--- a/plugins/virsh/tasks/cleanup.yml
+++ b/plugins/virsh/tasks/cleanup.yml
@@ -58,6 +58,7 @@
   ignore_errors: yes
 
 - name: cleanup known-hosts by ip
+  tags: networks
   known_hosts:
       state: absent
       name: "{{ item.1.ipaddr }}"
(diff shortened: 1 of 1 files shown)


mcornea:
Hi Dan,

We use GerritHub for reviews so could you send this patch to the following project: https://review.gerrithub.io/#/admin/projects/redhat-openstack/infrared ?

Thanks,
Marius

dprince:
@mcornea thanks. Just pushed some reviews to Gerrit.

PR #312, opened 2018-03-09
Set hosts user for all undercloud hosts

zzzeek:
When running infrared against an [undercloud] group that has more than one host, the Ansible "add_host"
directive by design will only run against one of those hosts, as it runs outside of the playbook loop.  Per Ansible documentation, add a "with" loop against all hosts in the "undercloud" group so that all undercloud entries in the inventory are updated.

Also I'm a red hatter but am not present in the redhat-openstack github group.

### DIFF
--- a/plugins/tripleo-undercloud/create_user.yml
+++ b/plugins/tripleo-undercloud/create_user.yml
@@ -84,8 +84,9 @@
 
       - name: Update hosts user
         add_host:
-            name: "{{ inventory_hostname }}"
+            name: "{{ hostvars[item]['inventory_hostname'] }}"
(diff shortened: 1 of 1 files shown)


obaranov:
@zzzeek, Thanks you for the fix. We are accepting contributions through the gerrithub (https://review.gerrithub.io/#/q/project:redhat-openstack/infrared). Can you please submit patch there? If you need any assistance or permissions, you can find me on IRC (#rhos-infrared (rh internal) or #infrared in freenode)

zzzeek:
OK great b.c. id rather be in that group

PR #329, opened 2018-06-07
Don't install virtualenv as user.

FedericoRessi:


obaranov:
@FedericoRessi, Can you please provide more info why you you need that change, cause we recently switched to the --user mode installation? 
Also looks like you're from redhat. Can you please submit patches to the gerrithub?  See details here: http://infrared.readthedocs.io/en/latest/contribute.html#red-hatters

PR #345, opened 2019-01-29
Make full string regex_replace compatible with Python 3

jistr:
Previously with Python 3 the replacement would glitch out and when
running IR with --prefix=jistr_test_ i'd get errors on provisioning
like:

"msg": "subnet jistr_test_external-subnetjistr_test_ not found"


jistr:
Closing in favor of gerrit review:

https://review.gerrithub.io/c/redhat-openstack/infrared/+/442589

PR #346, opened 2019-01-29
Allow setting router network in OpenStack provisioner

jistr:
Previously infrared just selected the first network from the list of
networks to create the router on. If that happened to be a network
which was IPv6 only, we'd get an error like:

Bad floatingip request: Network does not contain any IPv4 subnet.

This autoselection is still the default behavior, but the network
topology configuration can override it via a variable in the router
config.


jistr:
Thanks for +2, i was told to use gerrit so closing the PR in favor of gerrit review:

https://review.gerrithub.io/c/redhat-openstack/infrared/+/442591

PR #347, opened 2019-01-29
Fix specifying nics list to OpenStack provisioner

jistr:
Previously creating the BMC virtual machine resulted in error:

instance_id <ID> could not be found as device id on any ports

because the nics weren't passed into the module according to the
os_server docs. This is now fixed.


bregman-arie:
+2

PR #351, opened 2019-02-27
Go back to expanding /dev/sda1 by default.

lmiccini:
New-ish rhel8 images are back to the default setup with a single sda1 partition.

This commit removes the check and reverts to expanding /dev/sda1 by default.


lmiccini:
abandoning in favor of gerrithub.

PR #353, opened 2019-03-06
Update repository information for 13 and 14

knoha-rh:
repository information is a bit old and need an update for RHOSP13 and RHOSP14 which are already GA.


yprokule:
@aopincar shouldn't this go through [gerrit](https://github.com/redhat-openstack/infrared/pull/312#issuecomment-372245768)?
Now there is mismatch between [gerrit and github](https://review.gerrithub.io/#/c/redhat-openstack/infrared/+/447209/)

PR #354, opened 2019-03-07
READ-FIRST: USE GERRITHUB.IO FOR CONTRIBUTIONS

queria:
Read https://infrared.readthedocs.io/en/latest/contribute.html
before opening any Pull Requests.

TLDR:

- Red Hatters should send patches via gerrit (review.gerrithub.io)
  if You are one, DO NOT OPEN PULL REQUESTS please

- in other cases pull requests are OK
  (You will get bit more of human evaluation)

Change-Id: I8df121bf44ea203b3857de6b5f3451c574a37df7

### DIFF
--- a/USE-GERRITHUB-FOR-CONTRIBUTION
+++ b/USE-GERRITHUB-FOR-CONTRIBUTION
@@ -1,0 +1,1 @@
+https://review.gerrithub.io


hobbestigrou:
Sorry I have maybe a stupid question, but why the project I have two different ways to contribute? Gerrithub is open and use github. To add a new contributor on the gerrithub instance of the project, it's to add the new contributor in the rhosqeauto-contrib group. No specific internal dependency. I think this patch watch, that not a easy to have two different ways to contribute.

PR #365, opened 2019-08-23
Use file module instead of shell module

kajinamit:
It is not recommended to use shell module to remove files,
and we should use file module instead.


kajinamit:
Submitted the same patch to gerrithub

PR #374, opened 2019-11-07
Use python instead of python3 for the undercloud.conf.sample path. Resolves #373

### LINKED ISSUE #373: RHOSP 15 and 16 Unable to Copy Undercloud Sample Configuration
On RHOSP 16 Trunk (Train), the `/usr/share/python3-tripleoclient/` directory is actually named as `/usr/share/python-tripleoclient/`. This leads to a failure setting up the Undercloud.

```
$ ir tripleo-undercloud --version 16
...
TASK [copy default undercloud config] ********************************************************************************
fatal: [lshortundercloud-0]: FAILED! => {"changed": false, "msg": "Source /usr/share/python3-tripleoclient/undercloud.conf.sample not found"}
```

The logic in the code sets the directory to have "python3" instead of "python" for RHOSP >= 15.

ekultails:
For backwards compatiblity, TripleO no longer uses the python3 path.

### DIFF
--- a/plugins/tripleo-undercloud/configure.yml
+++ b/plugins/tripleo-undercloud/configure.yml
@@ -8,8 +8,6 @@
       - osp_release: "{{ install.version|openstack_release }}"
       - conf_instack: '/usr/share/instack-undercloud/undercloud.conf.sample'
       - conf_oooclient: '/usr/share/python-tripleoclient/undercloud.conf.sample'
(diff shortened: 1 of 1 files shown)


ekultails:
I have tested the patch on RHOSP 15 and 16 for deploying an Undercloud and it works in both cases.

PR #384, opened 2020-02-23
cdn_registery: added tag and repos for OSP16

nathancurry:


nathancurry:
not sure the best way to address the non-source pip install, so I just added a comment.  apparently relating to setuptools version from what I could find.

PR #391, opened 2020-05-23
Added note re: broken repo task for osp16.0

nathancurry:
The `get puddle url from the repos` task does not work for OSP 16.0, because target_install_version = 16.0, but the repo filename is `rhos-release-16.repo`.  Not sure whether should be fixed here or elsewhere. Not clear on how OSP 16.0 > 16.1 versioning is going to work going forward. Workaround for now is to manually edit repod_file variable.


nathancurry:
going to actually fix this in a minute

PR #392, opened 2020-06-03
Skip EC2 facts when deploying packstack

amoralej:
As reported in [1], in some CI environments metadata server takes longer
that expected to reply. This patch is creating a facter config file to
disable EC2 facts.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1842520

### DIFF
--- a/plugins/packstack/install.yml
+++ b/plugins/packstack/install.yml
@@ -33,6 +33,20 @@
             name: openstack-packstack
             state: present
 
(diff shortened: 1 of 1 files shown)


amoralej:
> o/ Alfredo, fyi infrared does not use github PRs but gerrithub/gerrit

Thanks! https://review.gerrithub.io/c/redhat-openstack/infrared/+/494897

Issue #411, opened 2021-12-21
Missing repos for 16.2


bshephar:
Issue:
While it is possible to deploy RHOSP16.2, if you do so with the `--cdn` option, it fails to lookup the repos since they are missing from the subscriptions dictionary:

Error:
```
TASK [cdn_registery : register to openstack repos for OSP >= 15] ***********************************************************************************************************************************************
fatal: [undercloud-0]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: {{ repo_list | map('regex_replace', '^(.*)$', '--enable=\\1') | join(' ') }}: {{ (subscription
s.common_8) + (subscriptions[(repo_version|float)] if not cdn_skip_openstack_repos | default(false) else []) + (subscriptions.ceph_4 if install.storage is defined and install.storage.backend is defined and in
stall.storage.backend == 'ceph' else []) }}: dict object has no element 16.2\n\nThe error appears to be in '/home/fedora/Documents/Code/OpenStack/infrared/.venv/lib/python3.10/site-packages/infrared/common/ro
les/cdn_registery/tasks/main.yml': line 17, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: register to openstack repos
 for OSP >= 15\n      ^ here\n"}
```

They will need to be added here:
https://github.com/redhat-openstack/infrared/blob/master/infrared/common/roles/cdn_registery/defaults/main.yml#L19

bshephar:
Apparently I don't have access to this repo on gerrithub. I'll push through the review once that access is sorted.

Or, if someone else wants to fix it. This should do it:

```
❯ git diff master
diff --git a/infrared/common/roles/cdn_registery/defaults/main.yml b/infrared/common/roles/cdn_registery/defaults/main.yml
index b1de3ad5..24721d2a 100644
--- a/infrared/common/roles/cdn_registery/defaults/main.yml
+++ b/infrared/common/roles/cdn_registery/defaults/main.yml
@@ -4,9 +4,9 @@ subscriptions:
         - rhel-7-server-extras-rpms
         - rhel-7-server-rh-common-rpms
     common_8:
-        - rhel-8-for-x86_64-baseos-rpms
-        - rhel-8-for-x86_64-appstream-rpms
-        - rhel-8-for-x86_64-highavailability-rpms
+        - rhel-8-for-x86_64-baseos-eus-rpms
[...]

PR #412, opened 2021-12-21
Fix repos for OSP16.2

bshephar:
RHOSP16.2 requires a 16.2 dictionary with
the relevant repos as we do for other
releases. Additionally, repo names are
slightly different and include 'eus'.

Resolves: #411


queria:
You should have access to gerrithub now, please try to use that for uploading changes.

PR #421, opened 2023-08-10
pullrequesttest1

wznoinsk:
let's see if this will be seen in devlake


fhubik:
It's seen in the Devlake, yet our PR's are useless to the MLTC metrics as it needs the merge date of real commit that wen't into the repo, not dummy one that is just "closed" at some point. I think we can close this because of that as this served its purpose.

PR #422, opened 2023-11-10
Create stf-connectors-cvp-osp17.1.yml

ayefimov-1:
New cloud template needed for new jenkins cvp-stf job so that stf-functional tests can run and pass


elfiesmelfie (commented):
IINM, this file needs to be moved to the plugins/tripleo-overcloud/vars/overcloud/templates directory.

You will see a bunch of other stf templates there.

